PR #1845, opened 2014-11-05
Hide Accordion`s H3 when there is only one

dennisschoenmakers:
I removed the header from the accordion when there is only one, but it feel too much like it is breaking the design.

image

Not yet ready to be merged yet! What do you think @antw and @ChaelKruip ?


ChaelKruip:
@dennisschoenmakers can we merge this one?

dennisschoenmakers:
> @dennisschoenmakers can we merge this one?

Yes we can!

PR #1869, opened 2015-03-30
Make scaling of scenario more obvious

dennisschoenmakers:
I've added an item to the menu

#### To consider:

- [ ] What if this is an already scaled scenario?
- [ ] Can we do this for all Areas?
- [ ] The flow is not logical now: you still have to save the scenario first. Could we add an option to the select list of the **Current* scenario?


dennisschoenmakers:
Include @antw to review and think about this.

Maybe we want to merge this now for the presentation today and improve on later?

ChaelKruip:
Thanks @dennisschoenmakers for making this happen on short notice and for the good questions posed!

> Maybe we want to merge this now for the presentation today and improve on later?

Merging it now for demonstration purposes. This week we can think about a more holistic workflow altogether.

Issue #1897, opened 2015-09-10
In the renewabilities chart, could we depict other fuels?


dennisschoenmakers:
e.g. when renewability percentage is 13.8%, could we indicate what the remainder consists of?

I understand that due to the complex definition of renewability this is **not** an easy task. :grin:

ChaelKruip:
Interesting point @dennisschoenmakers!

Do you mean something like this:
![20141209_topology_interface_v1_key](https://cloud.githubusercontent.com/assets/1303760/9783347/ed1ccc60-57a3-11e5-94f5-bf677d3aaca5.png)
?

But would that not raise the question what the subdivision of 'renewable' is (solar, wind, hydro, biomass etc?)

Then, if we would also make that subdivision would it still be clear what the renewable percentage is?

dennisschoenmakers:
> Do you mean something like this:

Yes, and I would also display the information on the left hand side (present year).

> But would that not raise the question what the subdivision of 'renewable' is (solar, wind, hydro, biomass etc?)

We could make a subdivision visible using shades of green?

![image](https://cloud.githubusercontent.com/assets/104830/9783531/38ec8562-57a5-11e5-9a27-01d601243e3e.png)

> Then, if we would also make that subdivision would it still be clear what the renewable percentage is?

That percentage will still be visible in the dashboard, right?

ChaelKruip:
> Yes, and I would also display the information on the left hand side (present year).

I was lazy :blush: 

> We could make a subdivision visible using shades of green?

Yes, but that would be a deviation from our normal color scheme e.g.,:
![energietransitiemodel_-_je_eigen_onafhankelijke__volledige_en_op_feiten_gebaseerde_energiescenario_](https://cloud.githubusercontent.com/assets/1303760/9783612/e9c2d904-57a5-11e5-9551-8bd7fced3d4e.png)

> That percentage will still be visible in the dashboard, right?

Good point!

dennisschoenmakers:
> that would be a deviation from our normal color scheme e.g.

True!

ChaelKruip:
I have kept the colours consistent with the rest of the ETM. What do you think @dennisschoenmakers @AlexanderWirtz is this enough of an improvement to justify inclusion in the upcoming deploy?
[...]

PR #1899, opened 2015-09-11
Translations

ChaelKruip:
Changed title for renewability chart.
Also added translation for efficiency slider of solar panels (not yet realised).

### DIFF
--- a/config/locales/en_input_elements.yml
+++ b/config/locales/en_input_elements.yml
@@ -59,6 +59,7 @@ en:
     costs_gas: "Natural gas"
     costs_oil: "Oil"
     costs_uranium: "Uranium"
+    efficiency_of_solar_pv: "Efficiency of solar PV"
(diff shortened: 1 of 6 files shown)


AlexanderWirtz:
See [my comments](https://github.com/quintel/etmodel/commit/413cdf5eb968d6d98d14359059bf72907f2aee70)

ChaelKruip:
> Should this not be: Percentage of renewable energy carriers in final demand?

I don't think so. The chart shows the percentage of **all** carriers in the energy use. The user is responsible to determine the fraction of those who are renewable. Showing only the renewable ones would not address the question raised in the issue (e.g., https://github.com/quintel/etmodel/issues/1897#issue-105753733).

I have decided not to give all renewables a similar color to not break consistency with the rest of the charts. Also see https://github.com/quintel/etmodel/issues/1897#issuecomment-139163888

PR #1903, opened 2015-09-17
Changing scaling view

grdw:


grdw:
@antw nevermind, this change is so small that I'll merge it myself.

PR #1947, opened 2015-10-29
Added slide, sliders and interface names in both English and Dutch. 

RobTerwel:
To do: add output elements names in English and Dutch and  add slides, sliders and charts on admin.

I translated 'LNG' to 'vloeibaar aardgas' and 'bio LNG' to 'vloeibaar biomethaan'. This does, however, have the consequence that 'regasified LNG' becomes ' hergast vloeibaar aardgas' - not quite my favourite group of three words. In my opinion, we could also just use the English terms and call 'regasified LNG' 'hergast LNG', for instance.

@jorisberkhout , does any of these options have your preference, or would you have a different suggestion?


jorisberkhout:
> I translated 'LNG' to 'vloeibaar aardgas' and 'bio LNG' to 'vloeibaar biomethaan'. 

If space allows, I would stick with the way CBS reports on this, i.e. 'vloeibaar aardgas (LNG)'. I would also change 'LNG' to 'Liquified Natural Gas (LNG)', again if space allows.

If texts become very long, I would stick to the acronym 'LNG' both in English as well as in Dutch.

> This does, however, have the consequence that 'regasified LNG' becomes ' hergast vloeibaar aardgas' - not quite my favourite group of three words. In my opinion, we could also just use the English terms and call 'regasified LNG' 'hergast LNG', for instance.

A quick search on Google for 'her-, ver- and hervergast' resulted in most hits for 'hervergast'. I would go for with the latter.

Could you make these changes and add them to this pull request, @RobTerwel ?

jorisberkhout:
> To do: add output elements names in English and Dutch

Could you add these to this pull request, @RobTerwel ?

Issue #1946, opened 2015-10-28
Naming of charts and carriers in charts: fix these


AlexanderWirtz:
Quite a few carriers and charts have been named confusingly and contain confusing carriers.`

### Final demand mekko needs carrier name update
- Gas should be network gas
- Bio-fuels should be biogenic_fuels

![image](https://cloud.githubusercontent.com/assets/308563/10787126/df8e93d0-7d6e-11e5-8ab8-834f03cc2c35.png)

- `network_gas` entry already exists
- Created new entries for: 

```
biogenic_fuels: "Biogene brandstoffen"
biogenic_fuels: "Biogenic fuels"
```

### Primary demand mekko needs carrier name update
- Biomass should be Solid biomass
- Gas should be Natural gas
- Biofuels should be Liquid biofuels

![image](https://cloud.githubusercontent.com/assets/308563/10787190/29e77f64-7d6f-11e5-9e23-fbad08374793.png)

- `natural_gas` entry already exists
- `bio_fuels` refers to Liquid biofuels
- Created new entries for: 

```
solid_biomass: "Vaste biomassa"
solid_biomass: "Solid biomass"
```

Still need to push this

### Rename Primary demand of biofuels chart
Should be 'Primary demand of biogenic energy carriers' and 'Primair gebruik van biogene brandstoffen' in Dutch

![image](https://cloud.githubusercontent.com/assets/308563/10787265/d97d2bc2-7d6f-11e5-8287-ba4aae0c8e83.png)

I have done this, but need to push this

### Electricity production
- Gas should be Natural gas
- Biomassa should be Solid biomass

![image](https://cloud.githubusercontent.com/assets/308563/10788200/4361d5be-7d76-11e5-84e1-db9a52a962e9.png)

### Import waterfall
- Gas should be Natural gas
- Biomass should be Biogenic ...

![image](https://cloud.githubusercontent.com/assets/308563/10788228/8ac45c4c-7d76-11e5-85d4-9448abd2054f.png)

### Geothermal is confusing in two charts as it is always zero
Residences:
![image](https://cloud.githubusercontent.com/assets/308563/10786716/01af6032-7d6c-11e5-8dcc-4552a86db2cf.png)

Other buildings:
![image](https://cloud.githubusercontent.com/assets/308563/10786723/0d5b2e0c-7d6c-11e5-8922-f98600fea531.png)

### Solar thermal is confusing as it reduces demand and is always zero
This confuses people.
[...]

PR #1951, opened 2015-11-10
Correction of chart labels, partially fixes #etmodel/issues/1946

RobTerwel:
New output elements in Dutch and English which partially fix https://github.com/quintel/etmodel/issues/1946 and add correct carrier names.


jorisberkhout:
Re-assigning to @AlexanderWirtz as he raised the issue.

PR #1977, opened 2016-01-08
Adding the correct wat to detect IE11

grdw:


grdw:
Since this was too small, I merged it myself.

Issue #1923, opened 2015-09-29
Group the charts by sector in the Demand


georgiasavvidou:
There are too many charts in the Demand and it would be more user friendly to group them by sector

![image](https://cloud.githubusercontent.com/assets/12493788/10164467/f43b7760-66b9-11e5-961f-6526bb0546eb.png)

ChaelKruip:
Moving to February deploy.

ChaelKruip:
Assigning @grdw

grdw:
Couple of questions. I have looked into it:

1. Is it important that the graphs stay in the same order? 
2. Should the demand charts be 'subgrouped' or are the demand charts split up into separate first level groups?

ChaelKruip:
> Is it important that the graphs stay in the same order?

No

> Should the demand charts be 'subgrouped' or are the demand charts split up into separate first level groups?

Subgroups I would think

grdw:
![screen shot 2016-01-19 at 3 09 55 pm](https://cloud.githubusercontent.com/assets/2676542/12420740/b90b7db8-bebe-11e5-9e7b-e329832eb879.png)

jorisberkhout:
> A preview

Looks good!

grdw:
Only downside is that you have to manually set the subgroups. I think that's possible within the admin panel but I'll have to take a look.

jorisberkhout:
As discussed is #2013 , let's pick this up after the deploy.

grdw:
Closing due to relevant PR in https://github.com/quintel/etmodel/pull/2220

grdw:
Only thing needed is filling in relevant sub-groups. Also we might want to use - as noted by @jorisberkhout - to use keys that can be translated. That latest part is not yet implemented but it's not that hard.

grdw:
Superseded by https://github.com/quintel/etmodel/issues/2612

PR #1984, opened 2016-01-19
Adding sub groups for output elements

grdw:
Fixes: #1923

### DIFF
--- a/app/views/output_elements/index.html.haml
+++ b/app/views/output_elements/index.html.haml
@@ -1,10 +1,11 @@
 #select_charts
   - @groups.each_pair do |group, charts|
-    - if charts.any?
-      %h1= t("chart_group.#{group.downcase}")
-      - charts.each do |chart|
-        %div.pick_chart{'data-chart_id' => chart.id, 'data-chart_holder' => @chart_holder}
-          %a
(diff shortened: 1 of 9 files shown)


app/views/output_elements/index.html.haml
```diff
@@ -1,10 +1,10 @@
...
-          %a
-            = image_tag "output_elements/#{chart.icon}", :size => "50x30"
-            %h2= t("output_elements.#{chart.key}").html_safe
+    %h1= t("chart_group.#{group.downcase}")
+    - if charts.is_a?(Array)
+      = render 'chart_preview', charts: charts
 
+    - elsif charts.is_a?(Hash)
+      - charts.each_pair do |sub_group, charts|
+        %h2= sub_group ? t("chart_sub_group.#{sub_group.downcase}") : 'Ungrouped'
```
antw: Tiny thing: Could we do away with the "Ungrouped" header? Just show the ungrouped charts at the top of the section without that `<h2>`? I don't think a visitor will care that those charts are ungrouped, so that header doesn't have much meaning to them.

grdw: Just reading this. I'll fix it :+1:

antw:
Thanks for the change to the "ungrouped" header. :+1:

I've merged this.

Issue #1847, opened 2014-11-13
The title for pop-up charts (for dashboard items) should be the same as their 'output element' title


ChaelKruip:
Related to https://github.com/quintel/etmodel/issues/1846.

The charts that pop-up whenever a dashboard item is clicked show a title defined in [e.g., en_dashboard_items.yml](https://github.com/quintel/etmodel/blob/master/config/locales/en_dashboard_items.yml). I think it would be better if the same title would be used as those described in [e.g., en_output_elements.yml](https://github.com/quintel/etmodel/blob/master/config/locales/en_output_elements.yml).

The direct reason for this request is that I want to use the same chart for the two CO2 dashboard items and I think it is confusing to the user that it now shows different titles depending on which dashboard item is used to open the chart.

antw:
I think you make a fair point, and have adjusted the local CO2 chart title to read "Total CO2 emissions". 

Changing the pop-up to instead show the output element title – as you suggest – is a better solution, but this will require us to make an explicit (database) reference between the constraint and the output element, and necessitates a couple of changes to the rendering of the pop-ups. Doing so might also allow us to eliminate [a couple of templates](https://github.com/quintel/etmodel/tree/18e8c49bfdb8b9f67151adfaa6a31637e9d301ff/app/views/constraints), which is a nice bonus.

This is certainly doable, but will have to wait until the _next_ deploy.

ChaelKruip:
> This is certainly doable, but will have to wait until the next deploy.

:+1:

ChaelKruip:
Moving to next deploy as discussed in the issue

jorisberkhout:
Assigning @grdw .

grdw:
I'm a bit confused about the relation between a `constraint` and an `output_element`. How would they relate?

antw:
A constraint would be better called a "dashboard item": one of the seven items which appears at the bottom of the page (energy use, CO2 emissions, energy imports etc). An output element is a chart or table, typically shown on the right-hand half of each page. If you click on a dashboard item, a chart or table ("output element") is normally shown.

[...]

PR #1990, opened 2016-01-22
Adding output element id to constraints

grdw:
Includes cleaning up the constraints.

Next to the now obsolete partials of constraints that existed in the database. I found two placeholder partials (`security_of_supply` and `score`) which I throwed away considering that they weren't rendered anyway. 

Fixes: #1847


antw:
Perfect! :+1:

Issue #1927, opened 2015-09-30
Improve investments table 2: decide on method to prevent double counting and implement 


AlexanderWirtz:
As @ChaelKruip and I agreed the improvements to the 'Investments for heat and electricity production' table have been split into two issues. This one deals with the double counting of heaters and water heaters (and cooling).

## Intro
The table is confusing, because the total costs listed at the bottom do not add up to the same number as the row items listed. Some technologies can be used for space heating and making hot water, and others even for cooling as well. Since these are separate converters, they are listed separately in the investments table, but that leads to an overestimation of actual investment costs. 

## Solution
John @ChaelKruip and I agreed for now to use the following method:

1. List each combi or heat pump technology only once, instead of two to three times
2. Use the investment costs number which is highest. As follows:
    1. Households: assume the investments in heat pumps (other than AC) for cooling are 0, IFF the corresponding heating / hot water heat pumps have investments > 0.
    1. Households: use the highest of the investments for space heating and hot water for each technology
    1. Buildings: use the highest of the investments for space heating and cooling for each technology
3. For total costs, just add up all costs of the line items in the table.

Also see #1926 

This issue is also relevant for the `Total cost` item in the dashboard, which currently calculates *average* costs for heat pumps. The impact of the suggested cost calculation method should be checked on these costs.

Issue #1926, opened 2015-09-30
Improve investments table 1: some converters need to be added


AlexanderWirtz:
As @ChaelKruip and I agreed the improvements to the 'Investments for heat and electricity production' table have been split into two issues. This one deals with converters that should be included in this table.

## Intro
This table costs us a lot of work, because users ask for the information in this table. Since the table is incorrect and unclear, we need to do extra work in Excel.

## Converters to include

The following converters are not (explicitly) found in the table, but should be. Some *are* included in the converter group `converters_in_investment_cost_table.gql ` but are not shown. Others are missing from said group and hence not shown in the table either. @ChaelKruip please check this.

Also, some of the converters I list below do not have sliders. We need to decide whether we want to keep the converters and add sliders or remove the converters.

```
- buildings_collective_geothermal 
- energy_chp_ultra_supercritical_lignite
- energy_heat_network_backup_heater_network_gas
- energy_heater_for_heat_network_crude_oil --> converter can be removed as far as I am concerned
- energy_heater_for_heat_network_geothermal --> I would like to keep this one
- energy_heater_for_heat_network_lignite
- households_flexibility_p2h_electricity
- other_burner_coal
- other_burner_crude_oil
- other_burner_network_gas
- other_burner_wood_pellets
```
The following industry converters may need a clearer name in the table so the user can tell them apart:
```
- industry_burner_coal
- industry_burner_crude_oil
- industry_burner_network_gas
- industry_burner_wood_pellets
- industry_chemicals_burner_coal
- industry_chemicals_burner_crude_oil
- industry_chemicals_burner_network_gas
- industry_chemicals_burner_wood_pellets
- industry_chp_ultra_supercritical_coal --> this has no slider and needs research
```
## Not to be included
The following should not be in this table or the converter group
```
- industry_aluminium_burner_network_gas
- industry_other_metals_burner_network_gas
- industry_steel_blastfurnace_burner_coal_gas
- industry_steel_electricfurnace_burner_network_gas
```

 Notifying @dennisschoenmakers and @jorisberkhout

ChaelKruip:
@AlexanderWirtz what about electric vehicles? They used to add no investment costs but now they will right?

ChaelKruip:
> The following should not be in this table or the converter group

Can you give an explanation for this @AlexanderWirtz?

RobTerwel:
> They used to add no investment costs but now they will right?

[...]

PR #1992, opened 2016-01-25
Investment table

ChaelKruip:
Changes to the investment table (see https://github.com/quintel/etmodel/issues/1926 and https://github.com/quintel/etmodel/issues/1927.


jorisberkhout:
I like it! :+1:

Issue #1921, opened 2015-09-29
Add a table view in dashboard


georgiasavvidou:
Consider to add a table view in the charts in the dashboard. This is useful for example in the case of 'Renewables' option in the dashboard where the share of 'Renewable gas' , 'Renewable heat' and other carriers in the Netherlands is so small that cannot seen by pointing at the chart.

![image](https://cloud.githubusercontent.com/assets/12493788/10164279/fc85df92-66b8-11e5-9415-2243eb3abf28.png)

RobTerwel:
Correct me if I'm wrong, but you can add a chart called `Renewability of the final demand' which seems the same. There is a table view option for that chart.

ChaelKruip:
> which seems the same

It is :smile: 

> There is a table view option for that chart.

Yes, but it would also be nice to have it for the pop-up chart (which is indeed the same).

RobTerwel:
>Yes, but it would also be nice to have it for the pop-up chart (which is indeed the same).

D'accord!

ChaelKruip:
Assigning @antw to do this if he feels like it...

ChaelKruip:
Moving to next deploy

jorisberkhout:
Assigning @grdw.

grdw:
Allright, I found how to turn a chart into a table. I found out where the chart is loaded but I don't get how this [fancybox](https://github.com/quintel/etmodel/blob/master/app/assets/javascripts/lib/views/constraint_view.coffee#L48-L57) ~~~~ connects to ~~~~~> [chart_list#load](https://github.com/quintel/etmodel/blob/master/app/assets/javascripts/lib/models/chart_list.coffee#L40-L101). According to some `console.log` statements I found out that there is something between them, but I can't figure out how to pass along more options, is that some Backbone wizardry? Could you maybe shine a little light on this matter @antw, if you know any of it. :innocent:

antw:
[...]

PR #1993, opened 2016-01-26
Turns renawables constraint pop-up into a table

grdw:
Fixes: #1921


antw:
I think the aim was for the *user* to be able to toggle between the chart view and table view [as in the normal charts](https://cloud.githubusercontent.com/assets/4383/12581920/b3df8fb4-c431-11e5-9812-03e029340314.png). This seems to force the renewability chart to be a table, with no option to show it as a chart.

Most charts can be changed into a table view, so I think it would be preferable to allow the user to be able to toggle between the two, rather than hard-coding one or the other. Maybe this option was already discussed and rejected?

ChaelKruip:
> I think the aim was for the user to be able to toggle between the chart view and table view as in the normal charts.

That's also what I thought...

grdw:
> Maybe this option was already discussed and rejected?

Not really. I sort of assumed that the chart was not practical due to the small numbers in the chart and that therefor the table view was prefered. I agree that a toggle would be nicer though.

grdw:
screen shot 2016-01-27 at 2 36 13 pm

I have now added the actions bar (with only the tabe action visible - since the other actions don't make any sense in a pop-up). This way you can toggle between the two.

jorisberkhout:
Assigning @dennisschoenmakers .

antw:
Is the `show_as` column still needed? I think we can probably drop that as the toggle button does everything we need.

dennisschoenmakers:
> 
image

To my taste this is a bit 'too subtle' and the button is just hanging there in the void. Can you try to make this clearer and more integrated with the design @grdw ?

AlexanderWirtz:
@grdw I agree. Perhaps actually turning it into a 'button' by framing it, or giving it some depth or something, it can be made clearer. This almost looks like a rendering error that will disappear when you hit refresh.

grdw:
screen shot 2016-02-01 at 11 26 30 am

Not sure about the alignment yet but it looks more button-like.

AlexanderWirtz:
much better. Alignment could be in the blue title ribbon above the chart, but the button is quite visible now at least

jorisberkhout:
I like it!

dennisschoenmakers:
> much better. Alignment could be in the blue title ribbon above the chart, but the button is quite visible now at least

Could you try to put it (right aligned perhaps?) in the light blue header? Maybe just a button without the primary color (so not blue, but grey) is also fine. This is not a 'primary' action (such as in a dialog to advance to the next page). 

And lastly: I think 'table view' is better, because that makes it clear(er) IMO that you're viewin the same information as a tabel, in stead of viewing (a?) table.

grdw:
screen shot 2016-02-01 at 12 24 00 pm

> Maybe just a button without the primary color (so not blue, but grey) is also fine. This is not a 'primary' action (such as in a dialog to advance to the next page). 

Hmm.. agree. partially. Don't really like the grey because blue matches great with blue.

screen shot 2016-02-01 at 12 25 47 pm

Anyone else have an opinion?

antw:
How about something in-between (smaller and plainer)?

screen shot 2016-02-01 at 11 33 35

dennisschoenmakers:
> How about something in-between (smaller and plainer)?

image

dennisschoenmakers:
Currently in the CO2 pop up chart the icon is not aligned properly:

image

And I am still missing [the icon](http://cl.ly/233u3l1e0a3P).

dennisschoenmakers:
And view should be **V**iew

grdw:
Like this:
screen shot 2016-02-05 at 11 26 46 am

dennisschoenmakers:
It's because of the 2 in CO2 I think. :smile:

grdw:
Some proof:

PR #2000, opened 2016-01-27
EDIT: Does not REMOVE the euro signs; instead changes it to €<x> bln/mln

grdw:
Can I get some virtual cake for hitting 2000?

---

As discussed in the issue.


dennisschoenmakers:
> Can I get some virtual cake for hitting 2000?

A muffin. Today at 16u! :smile:

PR #2003, opened 2016-01-28
Setting merit order warning for chart 164,163

grdw:


antw:
I wonder if we could make this more generic. I'm not a fan of hard-coding query or input names into any application because they can change. It's a pain if we have to check through every application whenever we rename something in ETSource.

Perhaps instead we could have a boolean `requires_merit_order` on the output_elements table? That would clean this up nicely, and make it so that non-programmers can set this on output elements that get added in the future.

ChaelKruip:
> Perhaps instead we could have a boolean requires_merit_order on the output_elements table?

:heart it!

jorisberkhout:
> Perhaps instead we could have a boolean requires_merit_order on the output_elements table?

Love it! @grdw, could you go ahead and implement this.

grdw:
I removed the hard-coded part and replaced it with a boolean as described. :+1:

antw:
:+1:

@grdw *When you have time*, please could you add checkbox/radio button form field for `requires_merit_order` [on the admin page](https://github.com/quintel/etmodel/blob/master/app/views/admin/output_elements/_form.html.haml)?

grdw:
> When you have time, please could you add checkbox/radio button form field for 

Yes, did it :+1:

antw:
Merged. Sorry for the delay. :blush:

Issue #1997, opened 2016-01-27
If the merit order is off, the load curves should not be downloadable


ChaelKruip:
I don't understand that they show information at all :confused:

antw:
> I don't understand that they show information at all

The merit order still runs in a very limited fashion when it is disabled; it needs to do so for the LOLE calculation.

jorisberkhout:
Assigning @grdw to get the buttons hidden when the MO is switched off.

These buttons that is!

<img width="465" alt="screen shot 2016-01-29 at 13 46 50" src="https://cloud.githubusercontent.com/assets/4669028/12675995/d5889160-c68e-11e5-87dd-ee89f08b3a43.png">

grdw:
Turned off:
![screen shot 2016-02-01 at 2 36 43 pm](https://cloud.githubusercontent.com/assets/2676542/12718820/488bdd02-c8f1-11e5-9d5a-a2edc64c7f4a.png)

Turned on:
![screen shot 2016-02-01 at 2 36 52 pm](https://cloud.githubusercontent.com/assets/2676542/12718819/488a441a-c8f1-11e5-9f34-6f2460ce35e2.png)

jorisberkhout:
Re-opening this as I don't see the download buttons anymore, even when the merit order switched on. @grdw, any ideas?

grdw:
> any ideas?

In all my honesty: not really. It's fine with me on `beta.pro`. Did you do anything special prior to toggling the merit order?

jorisberkhout:
I cannot reproduce it either. Closing.

PR #2006, opened 2016-02-01
Disable merit downloads buttons from merit-order page

grdw:
Fixes: https://github.com/quintel/etmodel/issues/1997

### DIFF
--- a/app/assets/javascripts/lib/views/base_chart_view.coffee
+++ b/app/assets/javascripts/lib/views/base_chart_view.coffee
@@ -33,6 +33,7 @@ class @BaseChartView extends Backbone.View
       @render()
 
   check_merit_enabled: =>
+    $(".merit-data-downloads").toggle(App.settings.merit_order_enabled())
     unless App.settings.merit_order_enabled()
       @container_node().html(
         $('<div>').html(I18n.t('wells.warning.merit')).addClass('well')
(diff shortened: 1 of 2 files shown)


ChaelKruip:
Now that we point the user explicitly to the merit-order toggle at the top of the page:
488bdd02-c8f1-11e5-9d5a-a2edc64c7f4a_png__1091x520_
I feel that we can remove the big button on the Merit Order slide itself. It's redundant and confusing IMO.

grdw:
@antw would this be as easy, as to remove InputElement with key `settings_enable_merit_order` ? Or is there more to it?

antw:
Aww, I always liked that big button. I like that there's something obvious beneath the merit order description, for turning it on and off, rather than it being hidden away in a settings menu.

If we must get rid of it, I think it should be disabled by setting the InputElement's "Dependant on" to "_always_hidden". That can be done on [the "edit" page for that element](http://beta.pro.et-model.com/admin/input_elements/614/edit). I expect the InputElement still needs to exist for the settings-menu checkbox to work.

Personally, I would prefer to keep it.

ChaelKruip:
> Personally, I would prefer to keep it.

I am very sympathetic to @antw's feelings of nostalgia :+1: 
It doesn't bother me that much anyways.

PR #2019, opened 2016-02-05
EV extra costs slider in front-end

RobTerwel:
Added a new sidebar item, slide, slider, chart and labels for the latter for the extra investment costs slider. See https://github.com/quintel/etsource/issues/800 .


jorisberkhout:
:+1:

PR #2059, opened 2016-02-10
Style flexibility order

antw:
Changes the style of the flexibility order "sortable" to better fit with other input elements in the application. I wanted to make it more obvious that the technologies are in order of preference and not just a list of technologies. I also wanted to make it clear to users that curtailment is not a sortable option, and is always the final technology.

Also fixed is a problem whereby technologies could be moved beneath curtailment, even though curtailment itself is not supposed to be sortable.

screen shot 2016-02-10 at 10 23 50

--

I tested in Chrome, Firefox and IE11. I don't have an IE9 or 10 to test with right now, but I don't expect any problems (except perhaps for them not showing the subtle background gradient behind the numbers).

Pinging @dennisschoenmakers @ChaelKruip @jorisberkhout @grdw for feedback. :grinning:

### DIFF
--- a/app/assets/javascripts/lib/views/base_chart_view.coffee
+++ b/app/assets/javascripts/lib/views/base_chart_view.coffee
@@ -110,19 +110,21 @@ class @BaseChartView extends Backbone.View
       base_url = "#{ App.scenario.url_path() }/flexibility_order/"
 
       $.ajax
-        url: "#{ base_url}get",
+        url: "#{ base_url }get",
         type: 'GET',
         success: (data) ->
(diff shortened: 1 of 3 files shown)


ChaelKruip:
I :heart it!

grdw:
:+1:  Looks good to me

dennisschoenmakers:
Awesome dude! Let's merge it!

jorisberkhout:
Love it!!

dennisschoenmakers:
Deployed to beta!

Issue #2083, opened 2016-03-17
Add a chart showing the hourly price data from merit


jorisberkhout:
#### Introduction

As described in more detail in #2081, we would like to give the user of the ETM more insight into the results of the merit order calculation. The hourly price curve contains very valuable information and deserves a separate chart.

#### Proposed implementation

- The price curve is readily available in ETEngine.
- It should be possible to switch between a view for the entire year and a view for a specific week in that year, just like in ETMoses.
- All profiles and related data are available in ETEngine.

As discussed with @AlexanderWirtz , @dennisschoenmakers and @ChaelKruip . Notifying @antw . I would be happy to hear your feedback!

Assigning @grdw . Talk to @ChaelKruip and me about the planning.

grdw:
@antw I have a few questions related to the price curves (and the participant load curves) from et-engine. I see that both are formatted as a CSV. The `ChartSerie` defaults to using gqueries to fetch the data from et-engine (?). 

1. Is there another way to grab that data from et-engine or is it just the CSV? 
2. If it's only to possible to fetch those CSV's should I expand on the: [https://github.com/quintel/etmodel/blob/master/app/assets/javascripts/lib/models/chart.coffee#L24-L27](chart serie switch) or should I create a whole new `OutputElementType`?

The CSV sometimes contains Infinity which is a bit of a problem (which I just saw).

antw:
> Is there another way to grab that data from et-engine or is it just the CSV?

It's currently CSV only, but I would suggest expanding on those actions so that they can return CSV *or* JSON. JSON is probably easier to handle in the ETM.

With that in mind, the existing ChartSerie/ChartSeries classes may not be of much use to you since they expect you to be using Gqueries. Perhaps you might be able to get somewhere by creating a new set of classes which are compatible with ChartSerie/Gquery (e.g. implementing things like `color` and `label`) but not limited to just a `present` and `future` value?

grdw:
> e.g. implementing things like color and label

I don't quiet understand where you are going at here (which label and color?). I'm guessing that these classes must do a call to `et-engine` to fetch the actual data (or is that supposed to be handled somehwere else in the backbone stack?). 

We could also make use of [d3 svg](https://github.com/mbostock/d3/wiki/CSV)?

antw:
[...]

PR #2097, opened 2016-04-25
Add merit order price curve

grdw:
Fixes: #2083

I kept the chart in CoffeeScript because of the `extent` syntax. I know that CoffeeScript does something funky compile-wise when that line occurs (I need to check what that is in some more detail later on). The date selector and the module that slices the load in bits is written in Javascript for that matter.


antw:
Some comments:

1. The chart doesn't seem to refresh. It should do so automatically when new Gquery results are received – provided the values returned have changed. As far as I can tell, a `refresh` method was added to the chart, so maybe a callback isn't firing?

  You can test this for yourself by visiting /scenario/supply/electricity/coal-plants and swapping "Pulverized coal" between 0 and max; ETEngine returns a new price curve (with different values), but the chart stays the same.

1. For the annual view, it would be nice to show only the month name. That is, "Jan", "Feb", ... instead of "Jan 01", "Feb 01". The "01" is a bit redundant.

2. The line itself could use `shape-rendering: crispEdges` to prevent it from looking blurry. But *only* on this chart; it would look wrong on charts with smooth lines.

  Default rendering:

  screen shot 2016-04-25 at 12 29 21

  `crispEdges`:

  screen shot 2016-04-25 at 12 29 17

3. There needs to be a thumbnail of the charts in assets/output_elements/merit_order_price_curve.jpg. This image would be shown here: 
  
  screen shot 2016-04-25 at 12 30 31

4. Needs NL translations for `price` and `merit_order_price_curve`. Would also be nice for the axis labels and date selector to be localised.

  screen shot 2016-04-25 at 12 33 13

grdw:
Thanks for the comments. I'll look into them.

jorisberkhout:
Thanks for your critical look at the chart, @antw !

grdw:
I fixed most of the feedback. I also made it so that whenever you select a week - lock the chart and start fiddling with the slider (in my test case Pulverized coal) the chart just updates that one week.

jorisberkhout:
Nice!

antw:
Looks good, refreshing works now.

I think `hourly_supply` and `hourly_flexibility` output elements can be removed form the migration (we can add those later when the charts are ready).

I just noticed that the EUR/MWh label has gone missing even though it is in [the screenshot](https://github.com/quintel/etmodel/pull/2097/commits/7c07a6902e530130f8a440a10ab5fb04e7084b5c#diff-a6aefc77c5013de003db070e73ccf818). Is this just me?

Regarding the date formatting, the ETM uses [i18n-js](https://github.com/fnando/i18n-js#date-formatting) so that we can do (a limited amount of) localisation of elements generated by JavaScript code. Perhaps that might be of use here?

antw:
I noticed something else (sorry). The vertical axis doesn't seem to re-scale when the values change. For example, here's a starting situation with max pulverized coal:

screen shot 2016-04-25 at 13 44 42
 
... and now reduced to 0 (max price is now ~325 eur):

screen shot 2016-04-25 at 13 45 44

grdw:
> I just noticed that the EUR/MWh label has gone missing even though it is in the screenshot. Is this just me?

screen shot 2016-04-25 at 14 44 53

It's also in my current screen. Maybe it's some margin thingy?

> Regarding the date formatting

That might be of some use here. :+1: 

> The vertical axis doesn't seem to re-scale

Crap.. I think I didn't really notice due to my boring charts. Moment please! :+1:

antw:
> It's also in my current screen. Maybe it's some margin thingy?

I don't think so. The label `<text>` element is empty. If I add some text to it, it appears:

screen shot 2016-04-25 at 13 49 50

I suspect the `unit` is missing for the `output_element`; I don't see it specified in the migration.

grdw:
> I don't see it specified in the migration.

That might be it. I added it manually after Joris told me what the correct unit was. I'll add it to the migration.

PR #2105, opened 2016-04-29
Merit charts

grdw:
Related and please merge first: https://github.com/quintel/etsource/pull/1108


antw:
The pricing and flexibility charts look good. I'm having an issue with the hourly supply chart though:

It loads with "Jan 01 - Jan 07" selected, but in fact displays the whole year (possibly without downsampling?), and the dashboard at the bottom of the page remains greyed out, even after the chart has loaded.

screen shot 2016-05-03 at 10 07 54

There appears to be a JS error when the data loads (which can be reproduced by changing the selected week):

screen shot 2016-05-03 at 10 11 27

I played with some of the supply and flexibility sliders (added a lot of wind production, and P2P batteries and EV). If you can't reproduce, let me know and I'll see if it's specific to my scenario.

--

I'm a bit concerned about the amount of data needed for the load chart; about 3.7MB for me. That's quite a lot to be transferring every time a user moves a slider, especially on a slow connection. However, it might be a bit better in production where the JSON will be compressed. Let's proceed but keep our eye on it; if the feature is popular we may have to figure out ways to send less data.

grdw:
> There appears to be a JS error when the data loads (which can be reproduced by changing the selected week):

I'm trying really hard to break it locally. Or should I use the beta version of et-engine?

antw:
Might be specific to my scenario then, I'll check it out. These are the inputs being used (it's all local, not on beta):

```yaml
--- !ruby/hash:ActiveSupport::HashWithIndifferentAccess
households_flexibility_p2p_electricity_market_penetration: 32.4
transport_car_using_electricity_availability: 38.9
number_of_energy_power_wind_turbine_coastal: 1315.4
number_of_energy_power_wind_turbine_inland: 16574.4
```

grdw:
I literally copy/pasted your inputs into my scenario and my chart looks as:

screen shot 2016-05-03 at 11 43 09

Date selector also works. I'll check Chrome, maybe it's a browser thing? (Highly unlikely but you never know).

antw:
Doesn't seem to be scenario or browser specific. I tried a brand new scenario in Firefox and got the same error:

screen shot 2016-05-03 at 10 41 25

antw:
> Date selector also works. I'll check Chrome, maybe it's a browser thing? (Highly unlikely but you never know).

Sorry, let me clarify. The problem I have is with the "Merit order hourly supply" chart; the flexibility and price charts work perfectly so I think it's specific to the hourly load chart.

grdw:
> Merit order hourly supply

That one also works just fine.. let me try to break it. Is the `updateData()` method line 85. Is that the `totalDemand` or the `chartData` throwing an error?

antw:
I think I found something. The loads for a technology may be an empty array, `[]`. If I change merit_transformator.js line 26 to:

```javascript
if (this.scope.stack && this.data.length) {
  // ...
```

... then the problem is solved. Perhaps it's trying to down-sample an empty array?

grdw:
Wait, do you have a total demand output element serie in your database for the hourly supply?

grdw:
@antw that's it. I just set the `is_target_line` attribute for the `total_demand` `output_element_serie` for merit order hourly supply to false and I got the same error you have 😄

PR #2121, opened 2016-05-23
Speeding up merit charts

grdw:


app/assets/javascripts/d3/merit_transformator.js
```diff
@@ -14,25 +14,17 @@ var MeritTransformator = (function () {
...
-                return this.data;
-            }
+            return this.data;
         }
     };
 
     function MeritTransformator (scope, data) {
         this.scope = scope;
-        this.data  = data;
+        this.data  = $.extend(true, [], data);
```
grdw: @antw this is now the slowest line by far now. Only thing is that we could maybe do a custom `$.extent` to maybe shave of 100ms or so? I'm a bit wondering if that'll help a lot though..

antw:
I fixed the `$.extend` bottleneck in fa1844be by waiting to clone the data until `transform()` is called, and then only doing a shallow clone of each object. The `values` will be a new object anyway (a result of calling `transformSerieValues`), so it's not necessary to clone the full length values array in the constructor.

#### Before

screen shot 2016-05-23 at 13 40 28

#### After

screen shot 2016-05-23 at 13 39 26

The biggest issue now is that when each Gquery result is retrieved, Backbone calls `_.isEqual` to determine if it needs to send callbacks notifying that a value has changed. There's not much we can do about that, and I think we've gone as far as we can with this.

@grdw Can you check out my changes, see what you think, and if you're happy then you can merge this.

grdw:
I agree. I already moved it from place to place to make it faster. Thanks for seeing yet another way 👍 

> There's not much we can do about that, and I think we've gone as far as we can with this.

Yes we did.

Issue #2114, opened 2016-05-13
When opening the new demand/flex charts together, the header of the lowest one becomes ugly and unresponsive


ChaelKruip:
The legend is plotted through the header and the header buttons don't work anymore:
<img width="692" alt="energietransitiemodel_-_je_eigen_onafhankelijke__volledige_en_op_feiten_gebaseerde_energiescenario_" src="https://cloud.githubusercontent.com/assets/1303760/15247424/72c316f8-1914-11e6-95d3-9a80b8c6038b.png">

PR #2135, opened 2016-05-25
Makes it possible to put multiple merit charts below each-other

grdw:
Fixes: #2114 

I don't think this will break any other tables/chart though but I just want to know why the height is statically defined?


antw:
> I just want to know why the height is statically defined?

To be honest, I have no idea. Thanks for changing this.

PR #2137, opened 2016-05-25
Translations for heat network mekko (174)

ChaelKruip:
Translations for Heat Network Mekko:

The translations are not ordered alphabetically (so you can edit them easily) but perhaps this should be done before merging this. I see this as a work-in-progress anyway.

See https://github.com/quintel/etsource/pull/1125 for the ETSource part


AlexanderWirtz:
Glad this chart exists now. I was struck by several things (in random order):

0. There is no slider for the `energy_heater_for_heat_network_geothermal` converter. I have been missing this in our work for IABR and GEA. It is a bit annoying that there is no way to produce a lot of geothermal heat outside of the sectors `housholds`, `buildings` and `agriculture`. I also cannot find whether we made a conscious decision to exclude it from the interface. I don't see why.
1. We badly need to impose some sort of order on the legend. Without alphabetic sorting, preferably by Demand and supply, it is illegible and useless to me. 
2. `energy_power_supercritical_waste_mix` is actually a chp, so should be renamed to: `energy_chp_supercritical_waste_mix`

Issue #2139, opened 2016-05-25
Sankey present year is still called 2012, should that be 2013?


dennisschoenmakers:
![image](https://cloud.githubusercontent.com/assets/104830/15544128/ee0152a8-2297-11e6-8e71-179d77a156bc.png)

jorisberkhout:
Good catch, @dennisschoenmakers . Simply renaming it to `Sankey energy flow diagram 2013` wouldn't do the trick, as we also support datasets that have present year 2012 (France, Germany, Spain, Poland, United Kingdom and NL2012) as well as present year 2011 (EU-27). Also, updating the chart title is apparently easily overlooked, given that the Dutch title of this chart still reads `Sankey diagram van energiegebruik 2010` :smile:

Could we instead rename this chart to `Sankey energy flow diagram (present)` and `Sankey diagram van energiegebruik (huidig)`? Any objections, @dennisschoenmakers , @AlexanderWirtz or @ChaelKruip ?

jorisberkhout:
Okay, after comparison with the titles of the "Present / Future energy imports" charts, I decided to rename the Sankeys to:

- Sankey diagram of present energy flows
- Sankey diagram of future energy flows
- Sankey diagram van huidig energiegebruik
- Sankey diagram van toekomstig energiegebruik

PR is on its way to @dennisschoenmakers  as soon as GitHub collaborates!

dennisschoenmakers:
> Could we instead rename this chart to Sankey energy flow diagram (present) and Sankey diagram van energiegebruik (huidig)? 

:+1: good idea.

PR #2140, opened 2016-05-25
Improved titles of merit order charts and Sankey's

jorisberkhout:
Closes #2137 and #2139 . Assigning @dennisschoenmakers , feel free to re-assign.


dennisschoenmakers:
We say 'use', but shouldn't we rather say 'load'?

And, what do you, @AlexanderWirtz, think is better?

1. Load of flexibility options **on hourly basis**
2. Load of flexibility options **per hour**

AlexanderWirtz:
> We say 'use', but shouldn't we rather say 'load'?
> And, what do you, @AlexanderWirtz, think is better?

@dennisschoenmakers 
Looking at the Dutch translations, I gather @jorisberkhout means 'use' in the sense of 'deployment' or 'utilization', not 'energy use'.

jorisberkhout:
> Looking at the Dutch translations, I gather @jorisberkhout means 'use' in the sense of 'deployment' or 'utilization', not 'energy use'.

Correct, I meant 'use' as the translation of the Dutch 'inzet'. 'usage' would be a possibility, too. What do you think, @dennisschoenmakers ?

I agree with the 'per hour' and will commit that change to this pull request.

dennisschoenmakers:
> @dennisschoenmakers 
Looking at the Dutch translations, I gather @jorisberkhout means 'use' in the sense of 'deployment' or 'utilization', not 'energy use'.

Agree!

dennisschoenmakers:
:+1: Excellent!

Issue #2133, opened 2016-05-24
Legend of Merit Order supply chart should exclude non-installed options 


ChaelKruip:
When zoomed-in it works:
![energy_transition_model_-_your_free__independent__comprehensive__fact-based_scenario_builder_](https://cloud.githubusercontent.com/assets/1303760/15509364/8defd08e-21d3-11e6-8dd7-067026726635.png)

But with the regular chart (in Chrome):
![energy_transition_model_-_your_free__independent__comprehensive__fact-based_scenario_builder_](https://cloud.githubusercontent.com/assets/1303760/15509381/9f0a3a76-21d3-11e6-83d1-97f0def97e5b.png)

PR #2153, opened 2016-06-17
Only draw legend items that add up to the merit

grdw:
- Includes updating legend when refreshing the chart

Fixes: #2133

### DIFF
--- a/app/assets/javascripts/d3/merit_order_hourly_supply.coffee
+++ b/app/assets/javascripts/d3/merit_order_hourly_supply.coffee
@@ -18,7 +18,7 @@ D3.merit_order_hourly_supply =
 
       @dateSelect.setVal(1)
 
-      @drawLegend(@series)
+      @drawLegend(@getLegendSeries())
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/d3/merit_order_hourly_supply.coffee
```diff
@@ -42,6 +42,14 @@ D3.merit_order_hourly_supply =
 
       Math.abs((d3.max(values) - d3.min(values)) / d3.sum(values))
 
+    getLegendSeries: ->
+      legendSeries = []
+      @series.forEach (serie) =>
+         if d3.max(serie.future_value()) > 0
```
antw: This could be...

```js
serie.future_value().find(function(value) { return value > 0 });
```

... which will return upon finding the first positive value (`max` will scan the whole array).

grdw: Thanks for the suggestion. I'll implement it as such.

Issue #2084, opened 2016-03-17
Add a table with statistical data on number and duration of excess electricity events


jorisberkhout:
As the charts that are described in #2081 and #2082 are potentially difficult to interpret, we would like to add a table with some key parameters about the number of times there is excess electricity and the duration of the excess events. I would suggest to keep this table very simple to start with the number of events with a minimum duration of 1, 2, 4, and 8 hours.

| minimum duration (h) | # of events |
| --- | --- |
| 1 | 100 |
| 2 | 75 |
| 4 | 5 |
| 8 | 2 |

Including @AlexanderWirtz , @dennisschoenmakers and @ChaelKruip.

jorisberkhout:
I'm struggling a little with this issue. I assumed that there is a load curve which contains the amount of excess electricity per hour (in MW) and which I could process to obtain the required data. I now realise that the flexibility feature in merit is implemented in a much smarter way.

@antw could you help me out here? I would basically like to have a method that I can call from a Gquery to fill the table above with data. This method would take 1 argument (`excess_duration`), i.e. an integer expressing the minimum number of consecutive hours with excess electricity. The method would then have to return another integer expressing the number of times throughout the entire year an excess electricity event lasts at least `excess_duration` hours.

Would you be able to estimate how much time it would cost to implement such a method, @antw ?

antw:
> I now realise that the flexibility feature in merit is implemented in a much smarter way.

Is that a polite way of saying "WTF was @antw thinking with this code"? 😆 

An excess load curve might be a nice addition, I'll look into that.

Would this be excess load before or after storage? i.e. if a time step has 50kWh of excess "always on" production, but 30kWh ends up being stored in batteries, do you want the curve to report an excess of 50kWh or of 20kWh?

AlexanderWirtz:
@jorisberkhout do we also want to report on the 'volume' of these excess events and not just on the duration?

jorisberkhout:
> Is that a polite way of saying "WTF was @antw thinking with this code"? 😆

Sorry, I overlooked your comment. I would of course never say or think anything like that!

jorisberkhout:
> @jorisberkhout do we also want to report on the 'volume' of these excess events and not just on the duration?

Good point, reporting the excess volume would be a nice addition to this table.

jorisberkhout:
[...]

PR #2158, opened 2016-06-24
Merit order/#2084

grdw:
Fixes: #2084


antw:
One small issue I noticed: if you mouseover the blue bars while they're animating to their new position, they get stuck and stop moving. I'm guessing this is due to the background colour animation when hovering over the bars.

This is most obvious when reducing supply such that there are no excess events:

screen shot 2016-06-29 at 11 24 33  screen shot 2016-06-29 at 11 31 09

The chart can look a bit empty when there are no events; some visitors might think the chart is broken. Would it be hard to add a short message saying there are no events? I'm thinking something like this:

screen shot 2016-06-29 at 11 34 09

grdw:
>  I'm guessing this is due to the background colour animation when hovering over the bars.

Yes. I think it's because D3 is not able to perform two transitions at the same time (it has to wait for one to finish). Considering the background color animation such a minor thing I'd consider removing the transition. We could instead - if it's desirable use CSS transitions instead of the D3 transitions. One downside to this, is that it won't work in [old IE versions](http://caniuse.com/#feat=css-transitions).

antw:
> We could instead - if it's desirable use CSS transitions instead of the D3 transitions. One downside to this, is that it won't work in old IE versions.

That's okay. I think we now target IE9 and above. I forget where we landed on that last time it was discussed... maybe IE10 and above?

Either way, it's okay to ignore IE compatibility when it's just eye candy. :grinning:

grdw:
Also includes #2160

antw:
TY!

Issue #2132, opened 2016-05-24
User should not be bothered with date-select for every chart


dennisschoenmakers:
Right now, you even have to select your week again for your zoomed-in chart. I think that's unfair. 😄 

But, when you have 3 charts open, you also have to select your new week for every chart: shouldn't we make that into one-select-to-select-them-all?

jorisberkhout:
Assigning @grdw to look into this.

grdw:
Do you mean that you want one select box or we could also synchronize the other select boxes? I'd rather prefer the latter. The reason behind this, is that if we have only one select box, it would be a bit of a hassle to find out where we would put that select box in the interface.

PR #2167, opened 2016-06-29
Synchronize all d3 date select boxes

grdw:
Fixes: #2132

### DIFF
--- a/app/assets/javascripts/d3/d3_chart_date_select.js
+++ b/app/assets/javascripts/d3/d3_chart_date_select.js
@@ -5,46 +5,62 @@ var D3ChartDateSelect = (function () {
         msInWeek = 604800000;
 
     function buildOption(i) {
-        var timeFormat = d3.time.format("%b %d"),
+        var nextIndex  = (i + 1),
+            timeFormat = d3.time.format("%b %d"),
             msOffset   = (msInWeek * i),
             start      = new Date(epoch.getDate() + msOffset),
-            end        = new Date(start.getDate() + msOffset + msInWeek - (msInWeek / 7));
+            end        = new Date(start.getDate() + msOffset + msInWeek - (msInWeek / 7)),
+            optionText = (timeFormat(start) + " - " + timeFormat(end));
 
-        this.weeks[i + 1] = [start, end];
+        this.weeks[nextIndex] = [start, end];
 
-        return $("<option/>").attr('value', i + 1).text(
-            timeFormat(start) + " - " + timeFormat(end)
-        );
+        return '<option value="' + nextIndex + '">' + optionText + '</option>';
     }
 
     function createOptions() {
-        var i;
-
-        this.selectBox.append($('<option value="0">Whole year</option>'));
+        var i,
+            options = '<option value="0">Whole year</option>';
 
         for (i = 0; i < 52; i += 1) {
-            this.selectBox.append(buildOption.call(this, i));
+            options += buildOption.call(this, i);
         }
+
+        return options;
+    }
+
+    function updateMeritChartsDate(s, val) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/d3/d3_chart_date_select.js
```diff
@@ -27,26 +27,35 @@ var D3ChartDateSelect = (function () {
         }
     }
 
+    function storeDate() {
+        window.globalDate = $(this).val();
```
antw: Instead of `window`, you might put this in the `App.settings` object instead.

```js
App.settings.set('merit_charts_date', value)
App.settings.get('merit_charts_date')
```

Then each chart can listen for changes to that attribute and update their own date setting as needed:

```js
App.settings.on('change:merit_charts_date', function(settings, newValue) {
  // ...
});
```

grdw:
> Instead of window, you might put this in the App.settings object instead.

Ok. Did that. 😄  

> Then each chart can listen for changes to that attribute and update their own date setting as needed:

That sounds a little redundant. Do you mean put this code in every chart so (`merit_order_price_curve`, `merit_order_hourly_load` and `merit_order_flexiblity`? I now [trigger](https://github.com/quintel/etmodel/blob/merit-order/%232132/app/assets/javascripts/d3/d3_chart_date_select.js#L30-L37) the `change.update` for every date select visible in the DOM when updating the `merit_charts_date` and is in my opinion the simplest, shortest code-wise way to achieving synchronized select boxes. I do agree that it's maybe not up to the date select to decide when a chart should refresh but hmmm...

antw:
> That sounds a little redundant. Do you mean put this code in every chart so (merit_order_price_curve, merit_order_hourly_load and merit_order_flexiblity?

This would be the way Backbone recommends doing things. The idea is that you have a model (`App.settings`) and a view (the date `<select>`). Views listen for changes in the model and update themselves accordingly, rather than telling other views to update.
         
Here's how it would look:

```diff
diff --git a/app/assets/javascripts/d3/d3_chart_date_select.js b/app/assets/javascripts/d3/d3_chart_date_select.js
index d03c874..1e0490e 100644
--- a/app/assets/javascripts/d3/d3_chart_date_select.js
+++ b/app/assets/javascripts/d3/d3_chart_date_select.js
@@ -27,15 +27,6 @@ var D3ChartDateSelect = (function () {
         }
     }

-    function storeDate() {
-        App.settings.set('merit_charts_date', $(this).val());
-
-        $("select.d3-chart-date-select")
-            .not($(this))
-            .val(App.settings.get('merit_charts_date'))
-            .trigger('change.update');
-    }
-
     D3ChartDateSelect.prototype = {
         selectBox: undefined,
         weeks: [ [epoch, new Date(1970, 11, 30) ] ],
@@ -44,9 +35,15 @@ var D3ChartDateSelect = (function () {
                 App.settings.get('merit_chart_date') || '1');

             this.selectBox = $("<select/>").addClass("d3-chart-date-select");
-            this.selectBox.on('change.store_date', storeDate);
-            this.selectBox.on('change.update',
-                updateChart || function () { return; });
+
+            this.selectBox.on('change', function() {
+                App.settings.set('merit_charts_date', $(this).val());
+            });
+
+            App.settings.on('change:merit_charts_date', function(s, val) {
+                this.selectBox.val(val);
+                updateChart && updateChart();
+            }.bind(this));

             createOptions.call(this);
```

I really don't have a strong preference in this case, but listening *for* changes is more the Backbone way.

grdw:
> I really don't have a strong preference in this case, but listening for changes is more the Backbone way.

Sounds like valid reasoning to me 👍 

I fixed it + cleaned up some of the code.

Issue #2136, opened 2016-05-25
Not able to put two merit charts next to each other (still a problem part II)


grdw:
What I did yesterday (why the issue #2127 got really confusing):

1. I started a new scenario
2. Added two of the merit charts (doesn't really matter which of the two)
3. Do a hard refresh (so Cmd+shift+R)

---

Errors in the console (model breaks):

<img width="1074" alt="screen shot 2016-05-24 at 17 31 41" src="https://cloud.githubusercontent.com/assets/2676542/15531712/16578830-225c-11e6-8312-ccbb26a4e6f5.png">

---

I saw that the a `serie` (almost at random) returned no values when doing this upon the initial `draw()` method however, `refresh()` is getting the correct values (because, `@dateSelect.setVal(1)` is triggering a `refresh()`)  which is causing the drawing to be correct. I'm a bit confused by this though because it looks as though some Ajax request is triggering the success method way before it's actually finished, or half of the data is being submitted. I couldn't really find out.

I'm wondering if you could reproduce this by any chance? @antw

dennisschoenmakers:
Assigned you @antw to have a **look** at this issue: perhaps you have got a clue where it's failing us? cc @jorisberkhout, @ChaelKruip and @AlexanderWirtz.

antw:
No `Priority` tag! Bliss! :innocent:

I can reproduce this. It seems adding a second chart triggers a second request to ETEngine; the queries required for the chart aren't included in the first call to ETE. When the response to the first request is received, the data for the *second chart* aren't there, and this error is triggered. The second response then arrives with all the gqueries needed to render the second chart.

It's been this way for a long time (`git blame` says 2012), but it seems like a silly limitation. The ETM *should* be able to request all the needed Gqueries with the first request.

dennisschoenmakers:
> I can reproduce this. It seems adding a second chart triggers a second request to ETEngine; the queries required for the chart aren't included in the first call to ETE. When the response to the first request is received, the data for the second chart isn't there, and this error is triggered. The second response then arrives with all the gqueries needed to render the second chart.

Good catch @antw! 

How come this behaviour isn't there (or noticeable) when you choose any other two charts?

antw:
I think the merit charts just assume that the results are there – and that they'll be arrays (when in fact the result is going to be something like `null` or `undefined`), and that's why the error messages appear in the console.

> How come this behaviour isn't there (or noticeable) when you choose any other two charts?

[...]

PR #2168, opened 2016-06-30
Merit order/#2136 multiple chart problem

grdw:
Fixes: #2136

### DIFF
--- a/app/assets/javascripts/lib/models/chart_list.coffee
+++ b/app/assets/javascripts/lib/models/chart_list.coffee
@@ -38,27 +38,33 @@ class @ChartList extends Backbone.Collection
   #
   # Returns the newly created chart object or false if something went wrong
   load: (chart_id, holder_id = null, options = {}) =>
-    @chart_requests = []
+    if @should_load_chart(chart_id, holder_id, options)
+      App.debug """Loading chart: ##{chart_id} in #{holder_id}
+                 #{window.location.origin}/admin/output_elements/#{chart_id}"""
 
-    # if the chart is already there...
-    current = @chart_in_holder holder_id
-    if current && current.get('chart_id') == chart_id
-      return false unless options.force
+      @chart_requests.push(@request_output_element(chart_id, holder_id, options))
+
+      @last()
+
+  should_load_chart: (chart_id, holder_id = null, options = {}) =>
+    should_load = true
+    current     = @chart_in_holder(holder_id)
+
+    if current && current.get('chart_id') == chart_id && !options.force
+      should_load = false
 
     # if we want to replace a locked chart...
     locked_charts = App.settings.get 'locked_charts'
+
     if locked_charts[holder_id]
       if options.force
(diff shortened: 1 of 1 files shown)


grdw:
@jorisberkhout could you test this out locally. And try to:

Add multiple charts, removing them, locking them while refreshing the page. Just try a couple of different charts and see what happens. If anything odd happens, let me know. Thanks! 👍

jorisberkhout:
> @jorisberkhout could you test this out locally

Will do later today.

jorisberkhout:
> @jorisberkhout could you test this out locally

Ok, I did and found two bugs.

### Bug 1: Legend for 'Merit order capacities...' chart is missing after some hard refreshes (this happens for other charts as well)
screen shot 2016-06-30 at 16 30 37

### Bug 2: Chart not shown (this happens for other charts as well) after some hard refreshes
screen shot 2016-06-30 at 16 30 47

Can you see if you can reproduce these, @grdw ? And fix them 😄

grdw:
> Bug 2: Chart not shown (this happens for other charts as well) after some hard refreshes

This one I can reproduce 👍 I'll try to check what's the problem. It's weird though that it also doesn't remember the order of the chart-holders 😩

jorisberkhout:
> This one I can reproduce 👍 I'll try to check what's the problem.

Thanks! I tried reproducing bug 1, but I can't. Should this ever pop up again, I will create a separate issue for this.

jorisberkhout:
Moved to `Deploy September 2016` milestone as it is not critical for this one.

antw:
> Moved to Deploy September 2016 milestone as it is not critical for this one.

Sorry, I haven't forgotten about this; just otherwise occupied, and wanted to take a closer look before merging. I'll do so now.

antw:
Looks like this solves the problem where multiple charts trigger extra requests. However it looks like the main request to ETEngine (`api/v3/scenarios/:id`) is delayed when you have multiple charts open:

| charts | img |
| :-: | --- |
| 1 | screen shot 2016-07-20 at 17 12 21 |
| 2 | screen shot 2016-07-20 at 17 12 01 |

That's not disastrous, since a delay is better than sending multiple requests, but it makes me curious why we have to wait to send the request if there is more than one chart? Where does the delay come form?

grdw:
> why we have to wait to send the request if there is more than one chart?

Hmm, good question. Don't know why exactly to be honest. I need to take another look for this. Also, I tried to make less of a mess of the `chart_list.coffee` model which might have caused this behavior.

jorisberkhout:
> Sorry, I haven't forgotten about this;

No worries, I just desperately wanted to close the other milestone 😄

grdw:
> That's not disastrous, since a delay is better than sending multiple requests, but it makes me curious why we have to wait to send the request if there is more than one chart? Where does the delay come form?

I can't really axplain the gap in-between either.. we might make a seprate issue of this and leave it be for the time being? What do you say @jorisberkhout?

Issue #2051, opened 2016-02-09
Charts look 'broken' when they're empty


grdw:
![screen shot 2016-02-09 at 3 50 15 pm](https://cloud.githubusercontent.com/assets/2676542/12919567/fd9f3c30-cf44-11e5-94e6-0fe973a559ca.png)

I think it would be more clear for users if there's an explanation why there's nothing here and display a 0 at the y-axis. 

@jorisberkhout should I assign myself? :angel:

jorisberkhout:
> @jorisberkhout should I assign myself? :angel:

Yes, good idea!

grdw:
I have an idea for implementing this which is really nice. I'll try to explain:

the `D3ChartView` receives two extra methods:

````coffeescript
class D3ChartView extends BaseChartView
  ...
  render: (force_redraw) =>
    ...
    @displayEmptyMessage()

  isEmpty: ->
    false

  displayEmptyMessage: ->
    if @isEmpty()
      # append some div-like element to the chart with a message
````

Than each subclass will be able to overwrite the `isEmpty()` method with their own implementation considering that there are probably multiple data-structures, values and attributes to validate before you can safely specify that a certain chart is considered to be empty.

grdw:
<img width="610" alt="screen shot 2016-09-09 at 15 23 33" src="https://cloud.githubusercontent.com/assets/2676542/18388309/693c62d4-76a1-11e6-8547-5efb3c5065d9.png">

I'm not sure about the message in the chart, if someone could maybe suggest something better. Also I'm open for opinions. 

When I come to think of it, when the chart is empty, why do we display the legend? It's a bit confusing. Also, we could remove the weird black line on the right side of the screen. What I'm sort of pointing towards is -- if the chart is empty, just display the message (in a bigger box) and leave it at that. Why render and calculate more things if the initial calculation returns 0?

ChaelKruip:
> When I come to think of it, when the chart is empty, why do we display the legend? It's a bit confusing. Also, we could remove the weird black line on the right side of the screen. What I'm sort of pointing towards is -- if the chart is empty, just display the message (in a bigger box) and leave it at that. Why render and calculate more things if the initial calculation is 0?

I like that.

grdw:
So, really really really minimal looking it would be:

[...]

PR #2222, opened 2016-09-12
Adding basics for displaying a message when the chart is empty

grdw:
Fixes: #2051


app/assets/javascripts/lib/views/d3_chart_view.coffee
```diff
@@ -26,8 +26,15 @@ class @D3ChartView extends BaseChartView
...
       @drawn = App.settings.merit_order_enabled()
 
+    @display_empty_message()
     @refresh()
 
+  is_empty: =>
+    false
+
+  display_empty_message: =>
+    EmptyChartMessage.display.call(this)
```
antw: Instead of `call(this)`, which I think is really ambiguous in `EmptyChartMessage`, what do you think of instead doing `EmptyChartMessage.display(this)`, and have `display` take an explicit parameter instead?

```js
var EmptyChartMessage = (function () {
        display: function (chart) {
            var container = $(chart.container_selector()),
                wrapper   = template.call(chart, container),
                isEmpty   = chart.is_empty();

                // ...
        }
}
```

antw: Same with `template`. I think having to `call` and `apply` is really confusing (and easy-to-forget), when simply passing in a parameter will do just as well...

app/assets/javascripts/lib/util/empty_chart_message.js
```diff
@@ -0,0 +1,39 @@
...
+        display: function () {
+            var container = $(this.container_selector()),
+                wrapper   = template.call(this, container),
+                isEmpty   = this.is_empty();
+
+            container.toggleClass("empty", isEmpty)
+
+            wrapper.height(this.height);
+
+            if (isEmpty) {
```
antw: This can be simplified as `wrapper.toggle(isEmpty)`.

app/assets/stylesheets/chart.sass
```diff
@@ -64,6 +64,23 @@
...
+          +background(linear-gradient(#fbfbfb, #f0f0f0))
+          border-bottom: 1px solid #bbb
+          width: 100%
+          text-align: center
+
+          p
+            padding-top: 135px
+            margin: 0
+
+          .anotation
```
antw: anotation → annotation

app/assets/javascripts/lib/util/empty_chart_message.js
grdw: > and have display take an explicit parameter instead?

I agree 👍 looks better.

> This can be simplified as wrapper.toggle(isEmpty).

Yes. 😅 didn't really spot it because the first code was a little bit more complicated than what it currently is.

Issue #2218, opened 2016-09-08
Investment table is broken on Beta


jorisberkhout:
Caused by the following failing gqueries.

<img width="703" alt="screen shot 2016-09-08 at 18 18 03" src="https://cloud.githubusercontent.com/assets/4669028/18357287/ab9485d2-75f0-11e6-9108-e50caa73c6ca.png">

jorisberkhout:
Closed with https://github.com/quintel/etmodel/pull/2234.

PR #2234, opened 2016-09-12
Fixed investment table

jorisberkhout:
I have:
- added the new burner types in the various industries
- removed the no longer existing burners
- added both P2G technologies

### DIFF
--- a/app/views/output_elements/tables/_chart_142.html.haml
+++ b/app/views/output_elements/tables/_chart_142.html.haml
@@ -11,7 +11,7 @@
       %th= t 'output_elements.tables.investments.investment'
       %th= t 'output_elements.tables.investments.investment'
       %th= t 'output_elements.tables.investments.difference'
-    %thead.subhead
+  %thead.subhead
     %th{:colspan => 4, :style => "text-align: center"}=t 'output_elements.tables.investments.agriculture'
   %tr.even
     %th= t "input_elements.agriculture_burner_crude_oil_share"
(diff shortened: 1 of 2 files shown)


app/views/output_elements/tables/_chart_142.html.haml
```diff
@@ -82,14 +87,14 @@
     %td{:data => {:gquery => "buildings_cooling_airconditioning_electricity_delta_in_investment_cost_table"}}
   %tr.odd
     %th= t "input_elements.buildings_cooling_collective_heatpump_water_water_ts_electricity_share"
-    %td.border-rgt{:data => {:gquery => "buildings_collective_heatpump_water_water_ts_electricity_present_in_investment_cost_table", :graph => :present}}
-    %td{:data => {:gquery => "buildings_collective_heatpump_water_water_ts_electricity_future_in_investment_cost_table"}}
-    %td{:data => {:gquery => "buildings_collective_heatpump_water_water_ts_electricity_delta_in_investment_cost_table"}}
+    %td.border-rgt{:data => {:gquery => "buildings_cooling_collective_heatpump_water_water_ts_electricity_present_in_investment_cost_table", :graph => :present}}
```
ChaelKruip: I think here the queries of the individual technologies are used instead of the ones that correct for double counting. This seems to be happening with a lot of the combi technologies below as well.

jorisberkhout:
This closes #2218.

PR #2242, opened 2016-09-13
Issues/2235

### LINKED ISSUE #2235: Labels are cut off by legend for Mekko's
The label in the Mekko below should read "Export and curtailment":

Can we plot long labels over multiple lines?

grdw:
Fixes #2235

### DIFF
--- a/app/assets/javascripts/d3/mekko.coffee
+++ b/app/assets/javascripts/d3/mekko.coffee
@@ -88,7 +88,7 @@ D3.mekko =
       legend_columns = if @carrier_list.length > 9 then 3 else 2
       legend_rows = @carrier_list.length / legend_columns
       legend_height = legend_rows * @legend_cell_height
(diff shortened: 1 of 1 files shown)


grdw:
Something odd happens at my localhost but I'm still seeing the overlap of labels in the Mekko chart...

antw:
Unfortunately...

screen shot 2016-09-14 at 09 26 51

I pushed a change to fix this by increasing the height of the `<svg>` element when the labels are too long and can't be wrapped.

antw:
> Something odd happens at my localhost but I'm still seeing the overlap of labels in the Mekko chart...

Are you able to reproduce that? If you can post a screenshot I'll try it as well.

grdw:
> Are you able to reproduce that? If you can post a screenshot I'll try it as well.

I'll make an issue.

PR #2292, opened 2016-11-03
update installation instructions

ploh:
update installation instructions

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -30,7 +30,7 @@ The ETM pro is released under the [MIT License](LICENSE.txt).
 
 ## Dependencies
 
-* Ruby 1.9.3 or Ruby 2.0
+* Ruby 1.9.3, 2.1.2 or 2.3
 * Mysql database server
 
 ## Installing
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -42,12 +42,12 @@ The ETM pro is released under the [MIT License](LICENSE.txt).
   cp -vn config/config.sample.yml config/config.yml
   cp -vn config/email.sample.yml config/email.yml
   ```
+  * Probably set "standalone" to `true` in "config/config.yml"
 
 * Run `bundle install` to install all the dependencies
-* Create local database with `rake db:create`
-* Fill the database with `rake db:reset`
+* Create and fill local database with `rake db:setup` and `RAILS_ENV=test rake db:setup`
```
grdw: > and `RAILS_ENV=test rake db:setup`

Considering that Etmodel uses Rails 4, this is no longer needed.

ploh: I know... but for me it did not work any other way - neither db:reset nor db:setup worked without the explicit environment. It is probably not worth spending time to investigate this, is it?

PR #2295, opened 2016-11-04
Explicitly specify ruby version and bump to 2.3.1

ploh:


dennisschoenmakers:
Hey guys, I am sure why we did this. What was the problem? Why is that reason/problem not documented here? We have been running for years without an explicit required Ruby version (with patch level)?

@jorisberkhout and @AlexanderWirtz apparently ran into major problems because of this.

PR #2310, opened 2016-12-02
New dashboard item: Share of renewable energy in households

### LINKED ISSUE #2299: New dashboard item: sustainability percentage for the household sector.
For Ameland, people want to see in the dashboard that the household sector is (e.g.) 100% sustainable rather than 40% of the *whole* energy system.

We would need to add a dashboard item that can be selected from the [**change menu**](https://cloud.githubusercontent.com/assets/104830/20217645/14f76adc-a821-11e6-8971-40cc57f67962.png).

In the change menu, a third item would be be added under *renewables*:

* Total share of renewable energy
* Total share of renewable electricity
* **Household share of renewable energy**

For your information, @ploh: 
* In the backend, a dashboard item is called a *constraint* :smirk:
* You will reed to write a gquery yourself in ETSource to calculate this, look at the other two ('total share of renewable energy' and 'total share of renewable electricity'). Test it locally, there is also a [**sandbox**](https://beta-engine.energytransitionmodel.com/data/latest/gqueries/test) available.
* Add NL and EN translations, no worries if the Dutch is off, we'll fix it.

ploh:
- This commit only contains translations and javascript
- Create the constraint (i.e. dashboard item) with:
  ```
  INSERT INTO `constraints` (`id`,`key`,`created_at`,`updated_at`,`gquery_key`,`group`,`position`,`disabled`,`output_element_id`)
  VALUES (18,'renewable_percentage_households',NOW(),NOW(),'dashboard_renewability_households','renewable',NULL,0,49);
  ```
- Also merge quintel/etsource#1200

Closes #2299

### DIFF
--- a/app/assets/javascripts/lib/views/constraint_view.coffee
+++ b/app/assets/javascripts/lib/views/constraint_view.coffee
@@ -78,6 +78,10 @@ class @ConstraintView extends Backbone.View
         Metric.round_number(result, 1)
       when 'renewable_percentage'
         Metric.ratio_as_percentage(result)
+      when 'renewable_percentage_households'
+        Metric.ratio_as_percentage(result)
+      when 'renewable_electricity_percentage'
+        Metric.ratio_as_percentage(result)
       when 'bio_footprint'
         formatted = "#{Metric.round_number(result, 1)}x"
         if App.settings.get('scaling')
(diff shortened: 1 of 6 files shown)


ploh:
> I adjusted the constraint title a little ("Renewables households" → "Households renewables")

:+1:

> Can a migration be added to do this? There is an example, here.

Of course it can (thx for the example). I did not do it because Dennis told me to do it manually - probably for me to learn about our databases. Or was there a specific reason why this should not be done with a migration, @dennisschoenmakers?

antw:
> Or was there a specific reason why this should not be done with a migration, @dennisschoenmakers?

When we deploy to production we often just copy database tables for semi-static data like constraints, so migrations aren't *really* needed. I still think they're useful for beta/staging and our local copies of the app (especially since raw SQL may depend in IDs which are different on the staging DB than your local copy).

PR #2312, opened 2016-12-07
Correct debug output

ploh:


antw:
I'm not sure I understand the reason for this change. The old link would take you to the admin page (which is useful if you want to understand or debug the chart), but the new link will just show the JSON used.

| Before | After |
| --- | --- |
| screen shot 2016-12-07 at 12 14 50 | screen shot 2016-12-07 at 12 16 10 |

If you want to see the JSON isn't the "Network" tab in Developer Tools a better option?

ploh:
> I'm not sure I understand the reason for this change. The old link would take you to the admin page (which is useful if you want to understand or debug the chart), but the new link will just show the JSON used.

I thought the debug output was just meant to log the URL that is subsequently gonna be fetched from the ETEngine server. But that would be a different domain anyways... which I did not notice because I am running everything locally... so my mistake. I will close this. And sorry for bothering you :slightly_smiling_face: 

> If you want to see the JSON isn't the "Network" tab in Developer Tools a better option?

Yes.

PR #2313, opened 2016-12-13
Local/derived dataset may be selected for new scenario

ploh:
Ref quintel/atlas@6cd3b7f15faf45569c5b66


app/helpers/layout_helper.rb
```diff
@@ -8,36 +8,74 @@ def back_to_model_link
...
-    label += " (#{ I18n.t('new') })" if opts[:test]
+    label = I18n.t("country_select.#{ code }", default: [code.to_sym, code.humanize])
+    label += " (#{ I18n.t('new') })" if area[:test]
 
-    content_tag :option, label.html_safe, :value => code, :selected => selected,
-      'data-earliest' => opts[:earliest] || 2013
+    content_tag :option, label.html_safe,
+      :value => code,
+      :selected => selected,
+      'data-earliest' => area[:analysis_year]
```
antw: I think [this should be `analysis_year + 1`](https://github.com/quintel/etmodel/pull/2313/files#diff-ccec77d3dd28c781f0ef06caa4e8c033L12). It doesn't make much sense to permit the `present` and the `future` to be the same year. I expect ETEngine will let you do it, but ETModel will produce confusing chart axes:

screen shot 2016-12-13 at 12 51 10

app/helpers/layout_helper.rb
```diff
@@ -8,36 +8,74 @@ def back_to_model_link
...
-      { area_code: area_code }
+      { area_code: area.area,
+        test: area.test?,
+        analysis_year: area.try(:analysis_year) || 2013 }
+    end
+  end
+
+  def derived_dataset_choices
+    Api::Area.all_by_area_code.sort_by { |area, _| area }.map do |_, area|
+      if area.base_dataset?
```
grdw: This is a bit of a werid check to determine if an area is a derived dataset. Would it be possible to turn this into something like: `area.is_derived?`. Also it would be nicer if you could do something like: 

````ruby
Api::Area.derived_datasets
````

And that all that logic would live in the model. I'm not sure if a helper is the correct place to put this code.

Comment (ploh) [inline app/helpers/layout_helper.rb]: You are right. I will improve this.

File: app/models/api/area.rb [hunk 3]
```diff
@@ -43,6 +43,10 @@ def self.all_by_area_code
     end
   end
 
+  def self.derived_datasets
+    all_by_area_code.sort_by { |area_code, _| area_code }.map { |_, area| area }.find_all { |area| area.is_derived? }
```
grdw: This line is a little to long.

You can add newlines:

````ruby
all_by_area_code
  .sort_by { |area_code, _| area_code }
  .map { |_, area| area }
  .find_all { |area| area.is_derived? }
````

.. but that's just partially solving it. I think this can be rewritten to:

````ruby
all_by_area_code.sort.map(&:last).find_all(&:is_derived?) # I think instead of `find_all` you can also use `select` but I think that they work similar on arrays?
````

ploh: I agree with your first version :+1:
The second is dangerous since it goes around the cache... but maybe we should just cache "all" instead of "all_by_area_code"... after thinking about it: We should stick with the indexed cache, otherwise the finder method has to do a linear search

ploh: btw:
> I think instead of `find_all` you can also use `select` but I think that they work similar on arrays?

`select` is just an alias for `find_all` :smirk:

grdw: For an `Array` it is 👼  but for a `Hash` it isn't as explained [here](http://stackoverflow.com/questions/20999192/is-find-all-and-select-the-same-thing).

ploh: Good to know. Thx :+1: 
I guess I should finally really stop using the [Pickaxe](https://www.amazon.com/Programming-Ruby-1-9-2-0-Programmers/dp/1937785491) as my reference doc because it is too outdated. What a pity!

PR #2315, opened 2016-12-16
Combine Co2 emissions charts

### LINKED ISSUE #2283: Pop-up chart CO2 could be a lot more interesting.
The current pop-up chart for CO2 (when you click the dashboard) is rather _flat_:

image

We already have a more in-dept analysis of the CO2-emissions per sector:

image

The pop-up chart would be a lot more informative when we would integrate the second chart above. I would prefer to keep the three columns rather, then we can keep the red line for the set target.

ploh:
- Integrate optional column for year 1990 into vertical stacked bar chart type
- Migrate Co2 emissions with 1990 chart to stacked bar chart type
  - Also show emissions per sector as in the old "Energetic Co2 emissions per sector" chart
- Delete the old Co2 per sector chart

Closes #2283


app/assets/javascripts/lib/models/chart_list.coffee
```diff
@@ -175,7 +181,23 @@ class @ChartList extends Backbone.Collection
...
-    $.when.apply(null, @chart_requests).done((d) =>
+    # The @chart_requests are jqXHR objects returned by $.ajax().
+    # Here, we want to wait until all Ajax requests have returned, no matter
+    # their success. But Promise.all (and similar constructs like $.when)
+    # has a fast-failure feature if one of the requests fails.
+    # Hence, we have to wrap the jqXHRs in new Promises which will
+    # always return successfully - after all, we are not interested in the
+    # success of the requests anyways (since success is handled directly in
+    # callbacks to the Ajax calls).
+    Promise.all(@chart_requests.map((request) ->
```
antw: Can you explain why the change from `$.when` to `Promise.all`? Promises [are not available in Internet Explorer](http://caniuse.com/#feat=promises) yet, so they cannot be used unless you also add a Promises polyfill library for older browsers.

ploh: I had to change the existing code anyways because the fast-failing feature of `$.when` prevented the successful calls from triggering the `done` function - or if we used `$.always` instead of `$.when`, the successful calls would not yet be back and therefore the charts would not be created yet (I tried this locally).
And then the JavaScript-native Promise-machinery seemed like the most natural choice to do this. But I am confident that I can change this to JQuery-Deferred-machinery.

ploh:
@antw Please have a look at my last commit avoiding the Promises.

antw:
Looks good @ploh! Thanks for making that change. :+1:

PR #2320, opened 2017-01-12
Set correct scaling attributes in session

grdw:
~~Related to https://github.com/quintel/etengine/pull/898~~

----

- [x] Don't save ScenarioScaling when using a local dataset

### DIFF
--- a/app/assets/javascripts/lib/models/input_element.coffee
+++ b/app/assets/javascripts/lib/models/input_element.coffee
@@ -12,7 +12,7 @@ class @InputElement extends Backbone.Model
   conversions: ->
     conversions = @get('conversions') or []
 
-    if App.settings.get('scaling') and Quantity.isSupported(@get('unit'))
+    if App.settings.get_scaling() and Quantity.isSupported(@get('unit'))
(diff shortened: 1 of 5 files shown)


app/models/api/area.rb
```diff
@@ -7,6 +7,11 @@ class Scaling < ActiveResource::Base
     self.site = Api::Area.site
   end
 
+  class Init < ActiveResource::Base
```
antw: Why are the initializer inputs sent with the area?

They're an implementation detail, and since there is (currently?) no plan to add a front-end editor for them it seems unnecessary to include them with the response.

grdw: > Why are the initializer inputs sent with the area?

Eh,, good question. I think they can be removed [here](https://github.com/quintel/etengine/blob/master/app/controllers/api/v3/areas_controller.rb) I guess. See this code as a patchy way of fixing the cookie bug.

grdw:
I fixed it [here](https://github.com/quintel/etengine/blob/local-datasets/app/controllers/api/v3/areas_controller.rb#L14)

PR #2333, opened 2017-01-25
Apply the correct scaling attributes when loading a scenario

grdw:
Based on: https://github.com/quintel/etengine/pull/903#issuecomment-274823170

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -84,6 +84,18 @@ protected
 
 private
 
+  def assign_scaling_attributes(params)
+    area = Api::Area.find_by_country_memoized(params[:area_code]) || Current.setting.area
+
+    if area.derived?
+      Current.setting.area_scaling = area.scaling.attributes
+    end
+
+    if params[:scaling_attribute]
+      Current.setting.scaling = Api::Scenario.scaling_from_params(params)
+    end
+  end
+
   def current_user_session
     return @current_user_session if defined?(@current_user_session)
     @current_user_session = UserSession.find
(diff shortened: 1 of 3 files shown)


antw:
Does this work for you? When I visit `scenarios/:id` (where `:id`) is a scenario based on a derived dataset, pressing "Open this scenario" sends the dataset scaling data to ETEngine:

```json
{
  "scenario": {
    "scenario_id": 606772,
    "scale": {
      "area_attribute": "number_of_residences",
      "value": 10000.0,
      "base_value": 7449298.0,
      "has_agriculture": null,
      "has_industry": null,
      "has_energy": null
    }
  }
}
```

Fortunately ETEngine rejects this (although only by chance):

```
Unpermitted parameters: base_value
Unpermitted parameters: base_value

# and ...

Mysql2::Error - Column 'has_agriculture' cannot be null:
  mysql2 (0.3.21) lib/mysql2/client.rb:80:in `block in query'
  mysql2 (0.3.21) lib/mysql2/client.rb:79:in `query'
  ...
```

This does not happen when creating a new scenario from the front page.

grdw:
Yes. This is working for me. Though, I do recognize that error. It means that ETEngine tries to create a `ScenarioScaling` which it shouldn't do because your area is already derived. But just to be clear:

You go to here:

screen shot 2017-01-25 at 17 10 59

And than you hit "Open this scenario"? 

Also on which branch is your `etengine`?

antw:
> Also on which branch is your etengine? (mine is on local-datasets-scaling-of-sliders)

My bad. I had changed back to "local-datasets" earlier today. Works fine with "local-datasets-scaling-of-sliders".

PR #2349, opened 2017-02-17
Assign correct scaling attribute when loading a scenario

grdw:
Fixes: https://github.com/quintel/etengine/issues/913

### DIFF
--- a/app/controllers/scenarios_controller.rb
+++ b/app/controllers/scenarios_controller.rb
@@ -108,9 +108,7 @@ class ScenariosController < ApplicationController
 
     Current.setting = Setting.load_from_scenario(@scenario)
 
-    if request.post?
-      assign_scaling_attributes(params)
-    end
+    assign_scaling_attributes(params)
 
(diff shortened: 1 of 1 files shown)


grdw:
@antw one clean-up question, is there a situation where we ever load the scenario through a `post` request? I couldn't find any use of it, but maybe you know somehow, by accident?

antw:
This is weird, because I'm pretty sure this was working when I tested it a couple of weeks ago. I remember you fixing it. 😕 

> one clean-up question, is there a situation where we ever load the scenario through a post request?

There is a `scenarios/load` route which goes to `ScenariosController#load` using `POST`: this is used by the scaling form. Ideally, this action should _always_ be a POST or PUT, and never a GET, since GET is meant to be idempotent (and starting a new scenario certainly isn't). That would be a nice way to clean this up.

Since that seems out-of-scope and I can't see any good reason for `assign_scaling_attributes` to have been limited only to `POST` requests, I think this may be merged.

PR #2380, opened 2017-04-18
Ref. `Re-style of title page` on Trello

dennisschoenmakers:
See Trello


ChaelKruip:
Nice work @dennisschoenmakers!

PR #2382, opened 2017-04-19
Fixes #2381

### LINKED ISSUE #2381: Current local scenarios cannot be exported to etmoses
Rather, the only option is to scale the scenario:

image

grdw:
Fixes #2381


antw:
[I don't think ETMoses will allow importing an unscaled scenario](https://github.com/quintel/etmoses/blob/460daf87663f531d78895edaea0c7b62ecee785b/app/models/import.rb#L88-L92). IMO it would be okay to remove the validation in ETMoses as it's quite unlikely anyone will manually enter an unscaled scenario ID.

ChaelKruip:
> IMO it would be okay to remove the validation in ETMoses as it's quite unlikely anyone will manually enter an unscaled scenario ID.

👍

grdw:
> IMO it would be okay to remove the validation in ETMoses as it's quite unlikely anyone will manually enter an unscaled scenario ID.

That validation is not there for that reason alone. It's also to prevent later errors. The 'scaled down to' value is used when importing. 

So I think we should pass a scaler object [from ETEngine](https://github.com/quintel/etengine/blob/master/app/models/api/v3/preset_presenter.rb#L33) when requesting a derived dataset.

ChaelKruip:
> The 'scaled down to' value is used when importing.

Perhaps (surely) I'm not understanding this fully but can't we pass a default scaling of '1' with unscaled scenario's?

grdw:
> The 'scaled down to' value is used when importing.

Just to clearify: what I mean with this are the `number_of_residences`.

antw:
> That validation is not there for that reason alone. It's also to prevent later errors. The 'scaled down to' value is used when importing.

Could we not do as we do with buildings and [instead use a Gquery for this purpose](https://github.com/quintel/etmoses/blob/460daf87663f531d78895edaea0c7b62ecee785b/app/models/import/buildings_builder.rb#L26)? It seems like it would be more consistent to use a query rather than the scaling value (even if they are the same number), especially since we might one day support scaling with a different attribute, such as land area, number of buildings, etc. ETE [does support that already](https://github.com/quintel/etengine/blob/master/app/models/scenario_scaling.rb#L17), even though it is no longer exposed to end-users.

Personally I don't feel it is correct to send a scaling (even if it is easy) since these are not scaled scenarios, and the scaling contains incorrect attributes. For example, for `drenthe`:

```js
{
  "area_code": "drenthe",
  "scaling": {
    "area_attribute": "number_of_residences",
    "base_value": 211902,
    "has_agriculture": false,
    "has_energy": true,
    "has_industry": false,
    "value": 211902
  }
}
```

* **`scaling.area_attribute`** is correct, but only by chance. Perhaps the derived dataset was based on a different attribute? In fairness, this is not currently supported, but perhaps in the future?

* **`scaling.has_agriculture`**, **`scaling.has_energy`**, **`scaling.has_industry`** are all default ScenarioScaling values. `has_agriculture` and `has_industry` are both incorrect. This does not affect the ETM interface *today* (since the area_scaling is used instead) but might lead to obscure bugs in the future.

I *think* we could avoid all this by adding a new `number_of_residences` query for use in ETMoses; I'm curious if you think that would be feasible? Does Moses need the number of households *before* it requests the Gqueries during the import?

grdw:
> Does Moses need the number of households before it requests the Gqueries during the import?

I don't think so; the `scaling_value` is used for 3 technology builders:

```
app/models/import/composite_builder.rb
app/models/import/house_builder.rb
app/models/import/resistive_water_heater_builder.rb
```

So I think it could be replaced with a gquery quiet easily I guess. Is the idea here to completely remove the `Scaling` module itself?

grdw:
Would it be as easy as this:

````gql
- query = AREA(number_of_residences)
- unit = number
````

Including @ChaelKruip

antw:
Looks good, thanks @grdw!

PR #2396, opened 2017-05-12
Fixes: #2393

### LINKED ISSUE #2393: Order of spikiness - by the looks of it - probably incorrect.

Related lines:

- Determining of spikiness for a serie of values: [here](https://github.com/quintel/etmodel/blob/master/app/assets/javascripts/d3/merit_order_hourly_supply.coffee#L38-L41)
- Sorting: [here](https://github.com/quintel/etmodel/blob/master/app/assets/javascripts/d3/merit_order_hourly_supply.coffee#L52-L53)

grdw:
Fixes: #2393


antw:
Looks better. :+1:

> Determining of spikiness for a serie of values: here

Another alternative would be to use the sample variance; I'm curious to know which people prefer (I have intentionally not identified which is which 😆 )...

> Sorting: here


I think part of the problem is that the series are ordered by their variance *throughout the entire year*, but only one week of data is shown which means that the sorting is not necessarily appropriate for the data which is *currently visible*. The charts look a lot better if you sort based only on the current week's data (but this does mean that the order of series can change when showing different weeks).

Here's what the first and second weeks of the year would look like if you sort based only on the visible data (these are from the same scenario as the *third* row in the table above).

| (1) | (2) | (3) |
| :---: | :---: | :---: |
| screen shot 2017-05-12 at 20 31 33 | screen shot 2017-05-12 at 20 32 20 | screen shot 2017-05-12 at 20 29 16 |
| screen shot 2017-05-12 at 20 31 48 | screen shot 2017-05-12 at 20 32 25 | screen shot 2017-05-12 at 20 30 20 |
| screen shot 2017-05-12 at 20 07 52 | screen shot 2017-05-12 at 20 08 24 |  screen shot 2017-05-12 at 20 07 21 |

@grdw @ChaelKruip @jorisberkhout Your thoughts?

antw:
After experimenting a bit more, I realised that another reason for things looking incorrect is that `sort()` is being used incorrectly. It should return -1 when `a < b`, `0` when they are equal, and 1 when `a > b`:

```coffee
@model.non_target_series().sort (a, b) =>
  varianceA = @getSpikiness(a)
  varianceB = @getSpikiness(b)

    return 0 if varianceA == varianceB
    if varianceA > varianceB then 1 else -1
```

Here's what it looks like with the fixed series sorting *and* calculating variance for only the currently visible data (as in the previous post):

| (1) | (2) | (3) |
| :---: | :---: | :---: |
| x1 | x2 | x3 |
| y1 | y2 | y3 |
| z1 | z2 | z3 |

grdw:
Looks good to me but could you maybe explain what the difference is between (1), (2) and (3)?

ChaelKruip:
My preference:

(1), (3), (2). Where (3) is a close second 😄

antw:
> Looks good to me but could you maybe explain what the difference is between (1), (2) and (3)?

I was hoping to get a sense of which of the three choices was most popular first, but okay. 😆 

1. This PR "spikiness" + [fixed sorting](https://github.com/quintel/etmodel/pull/2396/commits/79fe9fe8387fb83a12b15f0e8a82dcbe54d8c0b9)
2. [Sample variance](http://mathworld.wolfram.com/SampleVariance.html) + fixed sorting
3. Original version of  "spikiness" + fixed sorting

grdw:
Just a bit nitpicky maybe but we can we get rid of using underscore as a dependency for this fix? I'd prefer a vanilla Javascript solution as much as possible.

antw:
> Just a bit nitpicky maybe, but we can we get rid of using underscore js as a dependency for this fix? 

Sure, [this can be done](https://gist.github.com/antw/7a431dc941a6744172732669eda588bd), but help me understand why vanilla is preferred here. 😇 

The chart class depends on Backbone (as does most of the front-end), and Backbone depends on Underscore. Therefore we're never getting rid of Underscore with anything short of a rewrite of the front-end.

grdw:
> Therefore we're never getting rid of Underscore with anything short of a rewrite of the front-end.

I was just thinking; Quinn is dependent on underscore aswell. Just ignore my previous comment. Shall I merge this or should you? :)

antw:
> I was just thinking; Quinn is dependent on underscore aswell.

And jQuery. It's vintage. 😆 

> Shall I merge this or should you? :)

I haven't yet added the changes to sort only on the visible data. I'll add that shortly; today hopefully. If you'd like to check those changes I can pass this PR back to you once I've pushed them, if not I can merge it myself.

Issue #2415, opened 2017-06-19
IE9: Can't load a new scenario


grdw:
Starting a new scenario responds as though Etengine is offline. 

Debug info:

```
ApiGateway.update error [object arguments]
```

---

Starting an existing scenario responds partially (sliders load, charts don't show).

grdw:
Starting a new scenario responds as though Etengine is offline. Starting an existing scenario responds partially.

antw:
Deploy was ongoing. You probably caught it in the minute or two while the maintenance page wasn't up.

Seems fine now.

grdw:
I'm still seeing the same error on the console on the `beta-pro` server though.  I'm using IE9 (Windows 7) in VirtualBox.

grdw:
![screen shot 2017-06-19 at 16 12 35](https://user-images.githubusercontent.com/2676542/27289411-2486ee3c-550a-11e7-8923-1d8bd5e22849.png)

antw:
[It's this Airbrake error](https://quintel.airbrake.io/projects/32209/groups/1976140994971507083?tab=notice-detail). Seems to only affect IE<=9.

grdw:
Okay, I managed to squash this bug only to find lots of others 😢 . I'll create tickets for those bugs and hold those off  until further notice. I do not know how many IE9 users we have for the ETM but if it's worthwhile I'll try and fix them as soon as possible. 

Including @ChaelKruip

PR #2417, opened 2017-06-19
Fix formatting of parameters for the proxy api

grdw:
Related to severe problems in IE9

Fixes: #2415


antw:
Thanks @grdw. I've merged this and will send it to production right away.

I'm a little confused by this one since HTTParty says that the `:body` option can be something which responds to `#to_hash`... which is true of `ActionController::Parameters` (at least until Rails 5.1). 😕

PR #2515, opened 2017-10-25
Add insulation and heat pumps to the scenario report

antw:
@DorinevanderVlies I tried to update the NL report also, you might want to proofread it before merging to make sure I didn't mess up the meaning of the texts. TY!

See also: https://github.com/quintel/etsource/pull/1390

### DIFF
--- a/config/reports/main.en.liquid
+++ b/config/reports/main.en.liquid
@@ -843,6 +843,23 @@
     </p>
   {% endif %}
 
+  <!-- Inconsistency between heating options and insulation -->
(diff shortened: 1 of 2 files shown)


DorinevanderVlies:
> @DorinevanderVlies I tried to update the NL report also, you might want to proofread it before merging to make sure I didn't mess up the meaning of the texts. TY!

You're a fast leaner, two days in The Netherlands and you're able to write scenario reports in Dutch 😉

PR #2560, opened 2017-11-21
Require that input elements have a position

antw:
Validates that input elements have a numerical position assigned. Not having a position causes `sort_by(&:position)` to [throw an exception](https://quintel.airbrake.io/projects/32209/groups/2088489103767969068).


grdw:
Looks good to me! :shipit: 

---

Side note:
Do need to add, that people are adding a lot of these things directly in the database through SequelPro locally - circumventing the validations. 

Probably related: https://3.basecamp.com/3797409/buckets/4288547/messages/771412410

PR #2565, opened 2017-11-23
Issues/#2562

### LINKED ISSUE #2562: Change legends of all charts
As discussed during the last public-transport-pressure-cooker-meeting the order of the legend items should change for all charts. Now it is like this:
screen shot 2017-11-22 at 17 30 30
There are two things that need to be changed:

1.  Now the legend items are ordered from left to right, which makes it not clear for the user. It would be better if they would be ordered from top to bottom. 

2. Also it would be more intuitive to place the lowest item in the chart (the output_element_serie with position 1) also as the lowest item in the legend. 

Would it be possible to make these changes, @antw and @grdw ?

grdw:
Fixes #2562


app/assets/javascripts/d3/bezier.coffee
```diff
@@ -141,7 +141,7 @@ D3.bezier =
 
       @draw_legend
         svg: @svg
-        series: series
+        series: series.reverse()
```
antw: Is it right to make this change only for the bezier chart? I'm not keen on how this makes the bezier chart inconsistent with the others:

screen shot 2017-11-23 at 13 15 26

screen shot 2017-11-23 at 13 16 54

screen shot 2017-11-23 at 13 19 03

app/assets/javascripts/d3/merit_order_hourly_supply.coffee
```diff
@@ -32,8 +32,10 @@ D3.merit_order_hourly_supply =
         .offset("zero")
         .values((d) -> d.values)
 
-      @stackedData = @stack(@chartData[1..@chartData.length])
-      @totalDemand = [@chartData[0]]
+      lastId = @chartData.length - 1
+
+      @stackedData = @stack(@chartData[0...lastId])
+      @totalDemand = [@chartData[lastId]]
```
antw: Unfortunately, this needs to be repeated for *all* the Merit charts: dynamic demand curve, heat production, flexibility, etc...

Perhaps it's worth adding a general helper function somewhere (`Util`?) which will take an array of series and partition them appropriately: `{ targets: [targetSeries...], data: [dataSeries...] }`?

app/assets/javascripts/lib/views/d3_chart_view.coffee
```diff
@@ -115,30 +115,42 @@ class @D3ChartView extends BaseChartView
...
-    legend.append("span")
-      .text((d) ->
-        d.get('label') || I18n.t("output_element_series.#{d.get('key')}")
+    d3.selectAll(".legend-column")
+      .each((items) ->
+        scope = d3.select(this)
+
+        items.forEach((d) ->
+          item = scope.append('div')
+            .attr("class", () ->
```
antw: The idiomatic way to do this with d3.js [is with `classed`](https://github.com/d3/d3-selection/blob/master/README.md#selection_classed):

```js
item = scope.append('div')
  .classed('legend-item', true)
  .classed('hidden', d.get('skip'))
  .classed('target_line', d.get('is_target_line'))
  .on('click', @legend_click)
```

app/assets/javascripts/lib/views/d3_chart_view.coffee
```diff
@@ -115,30 +115,42 @@ class @D3ChartView extends BaseChartView
...
+              ["legend-item",
+                (if d.get('skip') then 'hidden' else ''),
+                (if d.get('is_target_line') then 'target_line' else '')
+              ].join(' ')
+            )
+            .on('click', @legend_click)
+
+          item.append("span")
+            .attr('class', 'rect')
+            .style("background-color", () => d.get 'color')
```
antw: I don't think `d.get('color')` needs to be wrapped in a function. Same for `text/label` below.

grdw:
@antw Fixed your remarks 👍

antw:
> The ability to toggle demand on the dynamic demand chart seems to have been broken by these changes. Maybe the `@totalDemand` variable is now referencing the wrong series?

This is still broken for me; is it working for you? Everything else looks good.

grdw:
> This is still broken for me; is it working for you? Everything else looks good.

Interesting. I think I broke it again, it did work. One moment.

antw:
Thanks @grdw; works perfectly now. Merged it! :+1:

Issue #2596, opened 2017-12-05
Remove custom chart type for `CO2 of biomass` chart


jorisberkhout:
I noticed that the vertical stacked bar chart has been updated with the possibility to include a column for the year `1990` which is relevant for the CO2 charts. The main CO2 chart has been updated accordingly, but the CO2 of biomass (chart 188) still uses a custom coffeescript format (`co2_emissions_biomass.coffee`). Let's get rid of this custom format by making it a vertical stacked bar making use of the `is 1990` feature.

Could you make this happen, @grdw ? Let me know of you have any questions.

Issue #2594, opened 2017-12-05
Legend order of chart #188 CO2 biomass


marliekeverweij:
I find the legend order of chart #188 (co2 biomass) a little funny, it looks a bit random (see figure below). @DorinevanderVlies , am I right that you have created this chart? Maybe you know why it's like this?

![screen shot 2017-12-05 at 13 53 41](https://user-images.githubusercontent.com/32833996/33608440-cd22821a-d9c4-11e7-9cbd-a0c36eeb7192.png)

![screen shot 2017-12-05 at 13 54 02](https://user-images.githubusercontent.com/32833996/33608444-cefb4928-d9c4-11e7-895c-77f7a39c9415.png)

ChaelKruip:
Perhaps we should hide all zero/small series *and* their legends?

DorinevanderVlies:
I think me and @jorisberkhout crated this chart together.

It becomes more clear when there is a significant amount of biomass in a scenario **and** the CO2 for biomass sliders are set:

![image](https://user-images.githubusercontent.com/19907658/33721403-37350efe-db67-11e7-880c-ed01a6c3ec58.png)

> Perhaps we should hide all zero/small series and their legends?
Great idea, maybe @grdw can help with this

jorisberkhout:
Please have a look at https://github.com/quintel/etmodel/issues/2596 before making any changes to this chart.

PR #2598, opened 2017-12-08
Issues/#2596 biomass co2

grdw:
This PR does a couple of things:

- It turns the co2_biomass chart into a vertical stacked bar chart
- Removes the co2_biomass chart as an `OutputElementType` from the database and as a separate coffeescript file
- Adds the possibility to hide perticular `OutputElementSeries` from the chart legend.

Fixes: #2596 
Fixes: #2594 

--

**MERGE FIRST: https://github.com/quintel/etsource/pull/1500**


jorisberkhout:
Looks good @grdw. While reviewing the result of changes in this PR, I realised that this chart is different in yet another.

The Gqueries that determine the `present` and `future` values are different, i.e. there is a Gquery called `climate_relevant_co2_biomass_gas_present` which returns a single value for the `present` instead of a value for both the `present` and the `future`. There is a separate Gquery called `climate_relevant_co2_biomass_gas_future` that only calculates a single value for the `future`. Comparing the two gqueries will make clear how they are different (they use a different set of sliders as their input).

Because of the changes in this PR, the single value for the `present` is now shown both in the `present` and the `future` column of the bar chart. I see two solutions:

1. Have the `..._present` gqueries return `0` for the `future` and vice versa
2. Combine the two gqueries `..._present` and `..._future` in a single one replicating the behaviour of the two individual gqueries

I don't know if this can work and would be happy to hear the opinions of @grdw and @antw .

antw:
> Adds the possibility to hide perticular OutputElementSeries from the chart legend.

Can we show this option in the admin interface (/admin/output_element_series/.../edit) – perhaps next to "Show at first" and "Is target line"?

antw:
> Combine the two gqueries ..._present and ..._future in a single one replicating the behaviour of the two individual gqueries

This seems like the correct option to me; we should avoid having separate `present` and `future` queries when there are meaningful values for both.

jorisberkhout:
> This seems like the correct option to me; we should avoid having separate present and future queries when there are meaningful values for both.

Cool, could you @antw or @grdw give a clue on how this could be done? I wouldn't know where to start. It's probably best if you have a look at the following gqueries as an example:
 [climate_relevant_co2_biomass_gas_present](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/vertical_bar_188_co2_emissions_biomass/climate_relevant_co2_biomass_gas_present.gql) and [climate_relevant_co2_biomass_gas_future](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/vertical_bar_188_co2_emissions_biomass/climate_relevant_co2_biomass_gas_future.gql).

grdw:
> Combine the two gqueries ..._present and ..._future in a single one replicating the behaviour of the two individual gqueries

This actually makes sense - for it removes the urge to hide certain legend items. 👼

grdw:
Okay done. The queries are now - reading the future or present query - depending on the 'scope':

https://github.com/quintel/etsource/pull/1500

This seems to work as expected:

*Future only:*

screen shot 2017-12-22 at 11 48 01

*Present only:*

screen shot 2017-12-22 at 11 49 06


*Both present and future:*

screen shot 2017-12-22 at 11 48 36

PR #2633, opened 2018-01-10
Highlight excess usage in carbon balance table

antw:
Highlights the carbon used in the "CO/CO2 available and used for power-to-kerosene" table when higher than the amount available.

This is a little quirky in that it requires that the gquery ends in "_used" and that there is a corresponding gquery also shown in the table ending in "_available". For example, to highlight the value for `direct_air_capture_co2_used` there must also be a value shown for `direct_air_capture_co2_available`. This *is* the case for all the values currently visible in the table.

screen shot 2018-01-10 at 14 43 57

screen shot 2018-01-10 at 14 44 09

screen shot 2018-01-10 at 14 47 31


jorisberkhout:
I like it! If you agree, can you merge this @marliekeverweij ?

PR #2689, opened 2018-03-01
Allow users to delete their account

antw:
The translations are from Google Translate and I don't expect them to be (anywhere near) perfect; would you mind double-checking them and adjusting as needed?

* Fixes the "edit account" page (previously would show information about creating an account), and adds a link to delete your account.
* Adds a confirmation page where a user must enter their password to authorise deletion.
* Confirming deletion removes the `User` and associated `SavedScenario` records, then signs the visitor out.

Required for [compliance with the GDPR](https://gdpr-info.eu/art-17-gdpr/).

---

screen shot 2018-03-01 at 12 47 28

---

screen shot 2018-03-01 at 12 47 37

### DIFF
--- a/app/assets/stylesheets/pages.sass
+++ b/app/assets/stylesheets/pages.sass
@@ -34,3 +34,81 @@
       line-height: 2em
     tr.even
       background-color: #F0F0F0
+
+.edit_account
+  width: 725px
+
+  h1, input[type=submit]
+    margin-left: 220px
+
+.edit_account, .delete_account
+  border: 1px solid #ddd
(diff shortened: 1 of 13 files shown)


grdw:
@antw Looking good! Everything works. I fixed some translations, if you could take a 2nd peak (I changed one English translation 🤓) that would be great.

The Dutch was fairly accurate for Google Translate 👍

antw:
> I fixed some translations, if you could take a 2nd peak (I changed one English translation)

Excellent thanks! Good catch on "with" → "wish".

Issue #2653, opened 2018-01-26
Google analytics


grdw:
- [x] Track page views / Track nagivation slides
- [x]  Track which sliders are touched
- [ ] Track which dashboard items are switched
- [x] Track which charts are loaded / added
- [ ] Track all options in settings:

![screen shot 2018-01-26 at 15 57 28](https://user-images.githubusercontent.com/2676542/35445360-a7f2ba12-02b1-11e8-876f-c8401cca4c0c.png)

**Already visible within the database:**

- Track if somebody turns off the merit order
- Track if scenario saved

PR #2690, opened 2018-03-01
Add basic google analytics metrics

grdw:
Fixes: #2653

### DIFF
--- a/app/assets/javascripts/lib/models/input_element.coffee
+++ b/app/assets/javascripts/lib/models/input_element.coffee
@@ -7,6 +7,7 @@ class @InputElement extends Backbone.Model
     @bind('change:user_value', @markDirty)
     @bind('change:user_value', @logUpdate)
     @bind('change:user_value', @update_collection)
+    @bind("change:user_value", App.analytics.changeInput)
     @bind('change:user_value', @additional_callbacks)
 
   conversions: ->
--- a/app/assets/javascripts/lib/app.coffee
+++ b/app/assets/javascripts/lib/app.coffee
@@ -12,6 +12,7 @@ class @AppView extends Backbone.View
     @scenario    = new Scenario()
     @router      = new Router()
     @merit_order = new MeritOrder(this)
+    @analytics   = new Analytics()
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/lib/models/input_element.coffee
```diff
@@ -27,6 +28,10 @@ class @InputElement extends Backbone.Model
 
     conversions
 
+  googleAnalytics: =>
+    if ga && ga.loaded
```
antw: For me, `ga.loaded` is always returning `undefined` so the analytics never run.

app/assets/javascripts/lib/models/input_element.coffee
```diff
@@ -6,6 +6,7 @@ class @InputElement extends Backbone.Model
       element: $("#input_element_#{@get('id')}")
     @bind('change:user_value', @markDirty)
     @bind('change:user_value', @logUpdate)
+    @bind('change:user_value', @googleAnalytics)
```
antw: To avoid binding this in every `InputElement`, this can be done in the `InputElementList` instead.

```js
class InputElementList extends Backbone.Collection {
  constructor() {
    // ...

    this.on('change:user_value', (inputEl) {
      if (ga && ga.loaded) {
        ga('send', 'event', 'slider-change', inputEl.key);
      }
    });
  }
```

app/assets/javascripts/lib/models/input_element.coffee
```diff
@@ -27,6 +28,10 @@ class @InputElement extends Backbone.Model
 
     conversions
 
+  googleAnalytics: =>
+    if ga && ga.loaded
+      ga('send', 'event', 'slider-change', "#{@get 'key'}")
```
antw: Nitpicking: Can this be made `input-change` instead of `slider-change`? ETModel tends to use the term "input" or "input element" as not all inputs are sliders (merit order switch, FCE switch, etc).

grdw:
I'm merging this to try something out really quick on beta. Will revert when it doesn't seem to work...

PR #2692, opened 2018-03-05
added gqueries for bezier final demand households

MartLubben:


MartLubben:
This does not entail the gqueries but the name of the chart.

PR #2712, opened 2018-03-08
Add a searchable select box for the dataset picker

grdw:
screen shot 2018-03-08 at 11 48 40

I kept the look and feel similar to the old one. It includes a search box.


grdw:
I synched both of the select boxes to look like `select2` boxes.

screen shot 2018-03-13 at 11 03 08
screen shot 2018-03-13 at 11 03 19

I've disabled the search functionality for the year selects. I'm going to see how hard it is to add flags, it doesn't seem to hard. So if you could give me a moment.

grdw:
screen shot 2018-03-19 at 12 10 15

This is the select box with an extended height + flags. I've pushed the changes to this branch. I'd love to hear what you think of this now @antw . Also - not to put a lot of pressure on this - but @jorisberkhout suggested that it would be nice if this could be merged for the Waterschappen project.

antw:
Looks good to me @grdw, I like it a lot. I made two tiny style tweaks:

* Removed the border radius on open select2 so that the select blends better with the options.
* Set a minimum width on the "other year" selection which avoids it being too narrow in Chrome (Firefox was fine).

| Before | After |
| :---: | :---: |
| screen shot 2018-03-19 at 12 04 54 | screen shot 2018-03-19 at 12 04 33 |

If you're happy with these changes, this can be merged as soon as you want.

PR #2716, opened 2018-03-09
Add basic google analytics metrics

grdw:
screen shot 2018-03-09 at 10 22 03

This seems to be working 👍

### DIFF
--- a/app/assets/javascripts/lib/models/input_element.coffee
+++ b/app/assets/javascripts/lib/models/input_element.coffee
@@ -7,6 +7,7 @@ class @InputElement extends Backbone.Model
     @bind('change:user_value', @markDirty)
     @bind('change:user_value', @logUpdate)
     @bind('change:user_value', @update_collection)
+    @bind("change:user_value", App.analytics.inputChanged)
     @bind('change:user_value', @additional_callbacks)
 
   conversions: ->
--- a/app/assets/javascripts/lib/models/input_element.coffee
+++ b/app/assets/javascripts/lib/models/input_element.coffee
@@ -7,8 +7,11 @@ class @InputElement extends Backbone.Model
     @bind('change:user_value', @markDirty)
     @bind('change:user_value', @logUpdate)
     @bind('change:user_value', @update_collection)
-    @bind("change:user_value", App.analytics.inputChanged)
     @bind('change:user_value', @additional_callbacks)
+
+    @bind('change:user_value', (input) =>
+      App.analytics.inputChanged(input.get('key'))
+    )
 
   conversions: ->
     conversions = @get('conversions') or []
--- a/app/assets/javascripts/lib/util/analytics.js
+++ b/app/assets/javascripts/lib/util/analytics.js
@@ -1,31 +1,36 @@
-var Analytics = (function () {
-    'use strict';
+window.Analytics = (function() {
+  'use strict';
 
-    function track(obj) {
-        if (window.ga) {
-            window.ga('send', 'event', obj.scope, obj.type, obj.data);
-        }
+  /**
+   * Create a new Analytics instance.
+   *
+   * @param {function} sender
+   *   A function which, when invoked, will send an event to the analytics
+   *   endpoint. May be an empty function if you want no events to be sent.
+   */
+  function Analytics(sender) {
+    this.sendEvent = sender || function() {};
+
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/lib/models/input_element.coffee
```diff
@@ -7,6 +7,7 @@ class @InputElement extends Backbone.Model
     @bind('change:user_value', @markDirty)
     @bind('change:user_value', @logUpdate)
     @bind('change:user_value', @update_collection)
+    @bind("change:user_value", App.analytics.changeInput)
```
grdw: @antw I'm not sure how to add this event to the `InputElementList` model. Also, why are the other event listeners on the `InputElement` and not on the List?

antw: > Also, why are the other event listeners on the InputElement and not on the List?

Good question, I don't have an answer for that. 😄

Perhaps just leave it as part of InputElement then? At least it's consistent with the other events.

app/assets/javascripts/lib/util/analytics.js
```diff
@@ -0,0 +1,31 @@
...
+    Analytics.prototype = {
+        chartAdded: function (key) {
+            track({
+                event: 'event',
+                type: 'chart-add',
+                data: key
+            });
+        },
+
+        changeInput: function (e) {
```
antw: Thinking about this a bit, should we really track *every* time a user touches an input? On the one hand it might show which sliders users spend the most time on (assuming that more changes = more time), but it might also drown out other interesting information like "how often is slider B used versus slider A"?

If we want to track that a user has touched a slider *one or more times* rather than logging an event for *each* change, [underscore's memoize](http://underscorejs.org/#memoize) can do the trick:

```js
function Analytics () {
  this.changeInput = _.memoize(this.changeInput);
}
```

If we do that, maybe the event needs to be called something else. "used input", "modified input", "touched input", ...?

@grdw @ChaelKruip Curious as to your thoughts.

---

Ditto for the charts. Memoizing would avoid the duplicate event being sent during the initial page load and would mean we only send an event the first time each chart is loaded.

app/assets/javascripts/lib/util/analytics.js
```diff
@@ -0,0 +1,31 @@
...
+    'use strict';
+
+    function track(obj) {
+        if (window.ga) {
+            window.ga('send', obj.event, obj.type, obj.data);
+        }
+    }
+
+    Analytics.prototype = {
+        chartAdded: function (key) {
```
antw: I've noticed that this gets fired every time a chart loads:

* Twice when the page loads for the default chart.
* Every time the user moves to a different slide and the default chart changes.

Maybe we'd be better splitting this into two separate events:

* "chart loaded" – what we have now
* "chart added" – fired when the visitor explicitly selects a new chart

app/assets/javascripts/lib/util/analytics.js
```diff
@@ -0,0 +1,31 @@
+var Analytics = (function () {
+    'use strict';
+
+    function track(obj) {
+        if (window.ga) {
+            window.ga('send', obj.event, obj.type, obj.data);
+        }
+    }
+
+    Analytics.prototype = {
```
antw: Nitpicking: the noun/verb order of event names in `Analytics` isn't consistent:

* **chart**Added
* change**Input**

app/assets/javascripts/lib/util/analytics.js
grdw: > Thinking about this a bit, should we really track every time a user touches an input?

So in other words - it's:

Checking which sliders are touched at all (regardless if the user drags it 1,2,3,4...400 times) **or** checking every time the slider is touched. I'm not sure what we want to track, so I don't really feel to strongly about this one.

app/assets/javascripts/lib/util/analytics.js
grdw: In this case I'm okay with just tracking every time it's 'added' and not how many times somebody loads the same chart.

app/assets/javascripts/lib/util/analytics.js
grdw: I actually discussed this last monday in short with @ChaelKruip and @jorisberkhout and it seems that only tracking which one's are used/touched is fine. As opposed to tracking everytime it has actually changed.

PR #2732, opened 2018-03-19
Reversed CO<sub>2</sub> to CO2

marliekeverweij:
I expected that it would work..... 

screen shot 2018-03-19 at 15 01 42


grdw:
This can work 👍 .. if you assign me or Anthony 👼

marliekeverweij:
I would be a nice-to-have to also be able to write the 2 of CO2 in subscript, just like the chart titles and slider names :) If you have some time, would it be possible to fix this, @grdw ?

grdw:
> If you have some time, would it be possible to fix this, @grdw ?

Maybe thursday would be possible. I'll note it down at least.

grdw:
Cheered to early, this is [not possible at the moment](https://github.com/fnando/i18n-js/issues/310) with our current implementation. We either need to rethink about how we draw legends to a chart (use a template of some sort - which takes time for it takes a lot of testing for all the different charts and legend layouts we have) or leave this be for now and live with the CO2 instead of CO<sub>2</sub>.

PR #2740, opened 2018-03-22
Move "Show CO2" button adjacent to "Start"

antw:
As discussed with @jorisberkhout on Slack, this PR moves the "Show CO2 footprint" button to be adjacent to the "Start" button, so as not to interrupt the flow of the start scenario form (Select region → select year → start).

The current button is both bold (immediately attracting the eye in spite of being the least important part of the form) and interrupts the flow of creating a new scenario.

1. User selects a region.
2. "Oh, what's this button?" \*click\*
3. "Hmm, I'm on a different page, but I wanted to create a scenario".
4. ???
5. ~~Profit.~~

| Before | After |
| --- | --- |
| screen shot 2018-03-22 at 19 05 55 | screen shot 2018-03-22 at 19 00 18 |

* Moves the button to the bottom of the form.
* I lightened the button slightly to match the select2 controls. Avoids the button standing out too much and attracting attention away from other parts of the form.
* There's a small border between the two buttons which I'm *hoping* pushes the eye towards "Start", which is what most people are going to want to press.

I'm open to other ideas if anyone has any; basically anything which removes the bold and moves the button away from the main flow of the form will be better.

I know we need to balance the wishes of a client, but this is perhaps the most important form in the entire application, and I feel the current placement and style of the button will confuse and risk reducing engagement.

---

A further improvement would be for the "Show CO2 footprint" button to pass through the selected year, so that the created scenario uses that instead of defaulting to 2040 every time.

### DIFF
--- a/app/assets/javascripts/intro.coffee
+++ b/app/assets/javascripts/intro.coffee
@@ -74,7 +74,7 @@ $ ->
     areaValue = $(this).val()
     selectedOption = $(this).find("option[value='" + areaValue + "']")
 
-    $("p.dataset_information a")
+    $(".dataset_information a")
       .attr("href", "/regions/" + areaValue)
 
   areaSelect.change(areaOnChange).change()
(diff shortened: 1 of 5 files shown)


jorisberkhout:
Look much better @antw ! Thanks for this.

Issue #2744, opened 2018-03-23
Create alert when hydrogen mekko cannot be shown


marliekeverweij:
When you open a new scenario no hydrogen is used or produced and therefore the hydrogen mekko looks like this:
![schermafbeelding 2018-03-23 om 20 42 02](https://user-images.githubusercontent.com/32833996/37850362-1f035828-2edb-11e8-9fb9-b18d4a9a6d22.png)

@ChaelKruip and I thought of the idea to show an alert here saying something like: 
_To view this information hydrogen needs to be produced or used somewhere in the model_

Just like now is done when the merit is not enabled:

![schermafbeelding 2018-03-23 om 20 42 43](https://user-images.githubusercontent.com/32833996/37850439-63b31576-2edb-11e8-884e-07e52847e5bf.png)

ChaelKruip:
@antw is this much work?

antw:
Something like this is supported already in the "excess events" chart when there are no excess events.

![screen shot 2018-03-29 at 14 15 53](https://user-images.githubusercontent.com/4383/38090833-b7f482bc-335b-11e8-8034-2ed42aaa97e6.png)

@grdw is probably in a better position to know how hard this is, but I can investigate as time  permits.

The only thing I'd add is that there are many charts which might benefit and I'm opposed to enabling it for only one or two charts or *types* of chart; it should be generic and supported on all of them. For example:

![screen shot 2018-03-29 at 14 14 45](https://user-images.githubusercontent.com/4383/38090975-121908da-335c-11e8-81ec-640e8ee8aa38.png)

grdw:
Every chart has the ability to overwrite the `D3ChartView`, `is_empty` method which should be filled with logic (which will have to return a boolean value) - also it needs a translation. That's all. If everything is correct it should render a message displaying that it's empty.  😄

Issue #2724, opened 2018-03-14
Legend of mekko charts


marliekeverweij:
For the hydrogen project I have created a mekko showing the demand and supply of the hydrogen network: 
![screen shot 2018-03-14 at 12 22 44](https://user-images.githubusercontent.com/32833996/37399798-2231f0cc-2783-11e8-949f-da03014d4d7c.png)

Recently we had changed the order of the legend items for charts in #2562 . Now the legend items go from L → R and from top → bottom.  

![screen shot 2018-03-14 at 12 25 33](https://user-images.githubusercontent.com/32833996/37399989-a94315fa-2783-11e8-8836-b4f567dd9e3d.png)

However, I see that this is not completely done for mekkos yet (only for vertical stacked bars and beziers I think). For mekkos L → R is done, but not top → bottom.

@grdw , could you maybe also change the order of the legend items for mekkos the same way as you did for the other charts? And I would also prefer if there are two columns with items and not three columns.

grdw:
![screen shot 2018-04-04 at 10 31 44](https://user-images.githubusercontent.com/2676542/38296998-63e5ea3e-37f3-11e8-87dc-3f85ea9b5a09.png)

Including @jorisberkhout in the abscence of @marliekeverweij  - This is what it's going to look like when there's no Hydrogen? I don't know what is not there 'energy-wise' so the translation might need some adjustments. 

Also - this can only be applied to **all** mekko charts. So therefor - mekko's that can become empty need a translation for when it does become empty.

michieldenhaan:
Looks good! Maybe change the text to 'There is no hydrogen supply and/or demand in this scenario'

grdw:
https://github.com/quintel/etmodel/pull/2757 would it be okay if I let you - @michieldenhaan - take care of this? 👍

AlexanderWirtz:
@grdw is @marliekeverweij 's orignial question clear?

grdw:
> is @marliekeverweij 's original question clear?

You mean the one about the actual ordering? I think so yes, and I also know I'm confusing this issue with another 😕 https://github.com/quintel/etmodel/issues/2744

grdw:
Back to the original issue:

> could you maybe also change the order of the legend items for mekkos the same way as you did for the other charts? And I would also prefer if there are two columns with items and not three columns.

I think the closest I am able to get - while keeping the original drawing of the legend items in tact - is like this:

[...]

PR #2757, opened 2018-04-04
Make the empty message appear for mekko charts

grdw:
Fixes: https://github.com/quintel/etmodel/issues/2744 & https://github.com/quintel/etmodel/issues/2724

### DIFF
--- a/app/assets/javascripts/d3/mekko.coffee
+++ b/app/assets/javascripts/d3/mekko.coffee
@@ -250,8 +250,6 @@ D3.mekko =
     display_legend: =>
       $(@container_selector()).find('.legend').remove()
 
-      legend_columns = if @carrier_list.length > 9 then 3 else 2
-
       series = _.filter(@carrier_list.models, (s) ->
         Math.abs(s.total_value()) > 1e-7
       )
@@ -260,7 +258,7 @@ D3.mekko =
         series: series
         width: @width
         vertical_offset: @series_height + @label_height
-        columns: legend_columns
+        columns: 2
 
     wrapLabels: =>
       label_height = @label_height
--- a/app/assets/javascripts/d3/mekko.coffee
+++ b/app/assets/javascripts/d3/mekko.coffee
@@ -250,15 +250,14 @@ D3.mekko =
     display_legend: =>
       $(@container_selector()).find('.legend').remove()
 
-      series = _.filter(@carrier_list.models, (s) ->
-        Math.abs(s.total_value()) > 1e-7
-      )
-
       @draw_legend
-        series: series
+        series: @prepare_legend_items()
         width: @width
         vertical_offset: @series_height + @label_height
         columns: 2
+
+    prepare_legend_items: =>
+      _.filter(@carrier_list.models, (s) -> Math.abs(s.total_value()) > 1e-7).reverse()
 
     wrapLabels: =>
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/d3/mekko.coffee
```diff
@@ -290,20 +298,6 @@ D3.mekko =
                      .text(word)
       )
 
-    # Increase the height of the SVG element to fit the largest unwrapped label.
-    fitHeightToLabels: ->
```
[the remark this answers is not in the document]
grdw: Well, it doesn't work with the `empty` combination. When `is_empty` switches from `true` to `false`, the height turns to 0 - which makes sense in the order the methods are executed. I thought this would no longer be needed but I guess it is. I'll see what I can do to make this work.

grdw:
> I was hoping to enable this for most – if not all – chart types, but I guess that's going to be a pain if custom translations have to be added for each one. It might also be more complex for charts which show future and present values.

Hmm.. we could overwrite the default `is_empty` and switch it with something like:

```coffeescript
# some pseudo-code
is_empty => 
  _.sum(@series.map((s) -> s.safe_future_value() + s.safe_present_value()) > 0
```
I'm wondering how many charts would break by actually doing this... :thinking:

grdw:
> Would it be possible to show a generic "empty" message if there isn't a custom translation for the chart? I can see someone adding a new mekko and forgetting the translations.

Yes. I think just removing the sub-line will do. That the default will just display: "There is no data to display for this chart".

app/assets/javascripts/d3/mekko.coffee
```diff
@@ -247,17 +250,23 @@ D3.mekko =
...
-
       @draw_legend
-        series: series
+        series: @prepare_legend_items()
         width: @width
         vertical_offset: @series_height + @label_height
-        columns: legend_columns
+        columns: 2
+
+    prepare_legend_items: =>
```
antw: Since Mekko charts only show future values, I don't think it's necessary to check the present value. I'm not sure why the legend list was changed from `_.filter(@carrier_list.models, (s) -> Math.abs(s.total_value()) > 1e-7)`; it's causing this because it enumerates through each series instead of each carrier:

screen shot 2018-04-11 at 09 12 43

grdw: The ideas was to reverse the output of the legend items. Funny that I didn't saw the multiple legend items. I'll fix it.

grdw:
Please hold off the merging. This breaks the label arrangement

grdw:
There's still one bug in this PR:

When you pick the hydrogen mekko:

screen shot 2018-04-11 at 11 02 33

And use one of the sliders on the left:

screen shot 2018-04-11 at 11 03 20

The chart height becomes 0. 

This is because of the code [here](https://github.com/quintel/etmodel/blob/master/app/assets/javascripts/d3/mekko.coffee#L305). When the chart is empty and it switches from 'Empty' to 'Non-empty'. The `<svg>` height is 0 at the moment the height is checked.

grdw:
@antw - I patched the problem described above with this https://github.com/quintel/etmodel/pull/2757/commits/c0707f85946ae214b4c2a975df6ce44ccab8de12 but feel free to suggest a smarter solution to this problem. 😅

Issue #2781, opened 2018-04-20
slider decimal places in local datasets


michieldenhaan:
In local datasets, some sliders have more decimal places than in country datasets / the step value specified in ETModel. This can be confusing: in the screenshot below the number of units of biogas CHPs can be interpreted as 31 thousand rather than 31.

![screen shot 2018-04-20 at 11 26 21](https://user-images.githubusercontent.com/32056448/39043652-418edaf2-448e-11e8-83a2-bc4abe7847ff.png)

In some cases this messes up the layout (the '2' of 'H2' is hidden behind the question mark):
![slack - quintel 2018-04-20 11-31-06](https://user-images.githubusercontent.com/32056448/39043700-61cb19ac-448e-11e8-8b7c-5d79b449c510.png)

grdw:
How would you like to see this work? My suggestion would be is to leave it to an input itself to determine how the max and min's are rounded. If we take for example [this input](https://github.com/quintel/etsource/blob/master/inputs/demand/households/households_demand/households_useful_demand_lighting.ad), I'd think that the rounding might work as:

```ruby
- query =
    UPDATE(
      V(households_useful_demand_light), preset_demand, USER_INPUT()
    )
- priority = 0
- max_value = 5.0
- min_value = -5.0
- start_value = 0.0
- step_value = 0.1
- unit = %
- update_period = future
- update_type = %y
#####
- round = 2
#####
```

However, if it will *always* be just simply 2 digits behind the dot I'd like to know, as it will make the solution to this problem far more easy.

michieldenhaan:
We already specify in ETModel how many decimal places we want to show. For local datasets, however, this setting is not respected. I don't think it's necessary to add an extra attribute, this issue would be solved if the sliders in local datasets display the same number of decimals as national datasets.

PR #2782, opened 2018-04-20
remove static width for slider units

grdw:
Part of #2781 

----

Removing static widths is always great. This is the difference before and after:

**Before:**
screen shot 2018-04-20 at 14 28 47

**After:**
screen shot 2018-04-20 at 14 28 43

----

**Before:**
screen shot 2018-04-20 at 14 29 22

**After:**
screen shot 2018-04-20 at 14 29 18

### DIFF
--- a/app/assets/stylesheets/slider.sass
+++ b/app/assets/stylesheets/slider.sass
@@ -84,7 +84,6 @@ $slider-box-height: 22px
 
   .output
     text-align: right
-    width: 73px
     white-space: nowrap
 
   .info-wrap


antw:
I appreciate the idea, but I don't think it really works out.

* It means sliders are not consistently positioned. This is a ugly annoyance in most cases...
  
  > screen shot 2018-04-23 at 11 34 48

  ... but really hurts the ability to quickly compare the relative positions in share groups:

  > screen shot 2018-04-23 at 11 36 01

* This happens when the width of the value changes:

  > kapture 2018-04-23 at 11 33 55

I think fixed widths are a must unless these issues can be resolved.

grdw:
@antw I was already thinking this would happen but I couldn't produce this. Thanks for doing this.

Issue #2795, opened 2018-05-29
Sankey not correctly displaying H2


AlexanderWirtz:
The Sankey currently does not seem to correctly show all H2 in a system. 

For example [this scenario](https://beta-pro.energytransitionmodel.com/scenarios/865980) that creates only hydrogen from excess electricity and imports hydrogen:
![image](https://user-images.githubusercontent.com/308563/40658150-93750d7a-634a-11e8-8d7a-ee2a7071dcc3.png)

There is only primary hydrogen (105 PJ of imported hydrogen is indeed primary, but the Sankey shows more, namely 290 PJ of primary hydrogen) and there is no electricity --> hydrogen visible in the Sankey.

Notifying @ChaelKruip and @michieldenhaan.
Discussed with @ChaelKruip to assign @marliekeverweij

marliekeverweij:
Oops! But I am not sure how we want hydrogen to be displayed in this sankey. 

Only primary carriers are shown in this sankey. Except for electricity and heat, they are produced from primary carriers in the middle of this sankey through electricity production and CHPs. Do we also want to display secundary hydrogen? Then we have to add many different conversion-blocks to the sankey I think (because now all produced hydrogen is being displayed in the left row but this actually should only be the imported hydrogen). For example we should add:
- conversion from electricity --> hydrogen with an `electrolyzer`-block right of `electricity production`
- conversion from gas --> hydrogen with an `SMR`-block underneath `CHPs`-block
- conversion from hydrogen -> electricity in an `gas turbine`-block

Is this how we want it? 

Then another question; how will we separate the produced hydrogen streams into the different sectors? Do we give each produced hydrogen stream the same split as shown in the 'demand/supply hydrogen mekko'?

Sorry lot of questions ;-)

There is also the easy option to only show the imported hydrogen and not the produced hydrogen from other sources, but I don't think that we want this.

AlexanderWirtz:
> Except for electricity and heat, they are produced from primary carriers in the middle of this sankey through electricity production and CHPs. Do we also want to display secundary hydrogen? 

I say yes: hydrogen plays a similar role in the system to electricity. It is interesting to know how much electricity goes into hydrogen production. This is probably complicated by the wish of Gasunie and GasTerra to make _dedicated_ hydrogen production primary hydrogen production and to show hydrogen produced from excess power as secondary hydrogen. Any other hydrogen is secondary anyway.

> For example we should add:
[...]

PR #2819, opened 2018-06-19
Update energy flow sankey

marliekeverweij:
See issue #2795 for information.


michieldenhaan:
Tried to fix as many as possible. Have to think some more about what to do with flexibility, synthetic kerosene and with micro chps/fuel cells and check whether the numbers add up now.

@antw I changed the queries going to `exports` but now the bottom part of the sankey is cut off. Do you know how to fix this?
I also added some code to hide entries with 0 energy in the label when you hover over a node (i.e. added ``if l.value() > 0.0 then``). It seems to work fine, but maybe you can double check if this is correct if you have time? 😇 


energietransitiemodel - je eigen onafhankelijke volledige en op feiten gebaseerde energiescenario httpetmodel test 2018-06-21 17-44-37

michieldenhaan:
I pushed my latest changes to ``update-energy-flow-sankey`` on both ETSource and ETModel.

Added flex as an extra column:
energietransitiemodel - je eigen onafhankelijke volledige en op feiten gebaseerde energiescenario httpetmodel test 2018-06-22 16-32-32

marliekeverweij:
We just had a meeting with @michieldenhaan, @ChaelKruip and @MartLubben and decided that we want to merge the current update of the sankey, because it is already a huge improvement compared to the one now on beta! And to make the sankey perfect would cost too much time for now. We will put the 'under construction' banner on the sankeys (present and future) since they are not finished and we will create a new issue on etmodel with the improvements still needed for this sankey. What do you think of that @antw?

The only thing that we would like to have fixed before merging is that part of the sankey is now cut off; but you are going to look into that right, @antw?

antw:
> you are going to look into that right, @antw?

Yes, I will find out why that happens.

antw:
> I will find out why that happens.

A fix has been pushed to this pull request. It should be ready-to-go.

screen shot 2018-06-26 at 14 07 07

PR #2864, opened 2018-09-11
added provinces translations - to production

michieldenhaan:
@antw if you have time, could you be so kind to deploy this to production (and flush the cache of both beta and pro)?


antw:
> deploy this to production (and flush the cache of both beta and pro)?

Done.

PR #2908, opened 2018-11-21
Add "related hourly chart" button

antw:
Needs [a couple of translations here](https://github.com/quintel/etmodel/blob/show-hourly-chart/config/locales/nl_output_elements.yml#L59-L60).

For the moment the button is shown as part of the chart header. It's not that obvious, but a larger button below the chart felt very out of place.

kapture 2018-11-21 at 15 26 00

A chart can be made to link to another by setting the "Related output element" in the admin section:

screen shot 2018-11-21 at 15 31 32


michieldenhaan:
Cool! I've added the translations here: https://github.com/quintel/etmodel/commit/27ac9ec20cc5710b37b3673f74ff7a697c9a5d30

PR #2909, opened 2018-11-28
Add "results" tip/popup

antw:
[Only one translation is needed](https://github.com/quintel/etmodel/blob/57059a61dc46855946ad5cc6f8efe60e35c73603/config/locales/nl_etm.yml#L87), if you'd be so kind. 🤗

---

screen shot 2018-11-28 at 15 05 14

60 seconds after a scenario loads, a small pop-up will appear in the sidebar directing the user's attention to the results tab. I wanted to show it after the user had moved a slider or two, but this would become messy; I think 60 seconds should be long enough for a visitor to start using the ETM without it being so long that they never see it.

The user may temporarily or permanently hide the popup. Temporarily hiding will prevent the popup from reappearing until the user starts a new scenario.

Permanently hiding it depends on whether the visitor is signed in:

* **Guests:** the tip remains hidden until the visitor clears their cookies, or we delete user sessions during a deploy.
* **Signed-in visitors:** The preference is saved in the database and the popup will never be shown again while the user is signed in.

### DIFF
--- a/app/assets/javascripts/lib/views/results_tip_view.js
+++ b/app/assets/javascripts/lib/views/results_tip_view.js
@@ -1,0 +1,98 @@
+(function() {
+  var closeButton = _.template(
+    '<li>' +
+      '<button class="<%= cssClass %>">' +
+        '<% if (showIcon) { %>' +
+          '<span class="fa fa-times"></span> ' +
+        '<% } %>' +
+        '<%- message %>' +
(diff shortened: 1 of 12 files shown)


ChaelKruip:
> Only one translation is needed,

Done!

ChaelKruip:
Nice work @antw !

PR #2920, opened 2019-01-02
Change "related"/"previous" chart buttons into toggles

ChaelKruip:
@marliekeverweij can you test this locally and merge it if you find no problems?


marliekeverweij:
Yes I will do this today!

PR #2922, opened 2019-01-02
Use icon for demand response also for fuel prices

marliekeverweij:
Currently the demand response sidebar item does not have an icon. I propose to use the same one as for fuel prices as that also kind of fits to demand response, this one: fuel_costs

Or do you have another idea?


ChaelKruip:
> Or do you have another idea?

What do you think about this one?

demand_response

ChaelKruip:
> How could I put it in the right format?

I used Preview.

demand_response_bw

PR #2941, opened 2019-02-01
finalized scenario reports

MartLubben:


MartLubben:
And this etsource update is needed: https://github.com/quintel/etsource/pull/1893

PR #2952, opened 2019-02-07
Add new landing page

antw:
This is the pull request for the new landing page for the ETM.

@marliekeverweij The translations needed can be found in [config/locales/nl_intro.yml](https://github.com/quintel/etmodel/blob/simple-root/config/locales/nl_intro.yml)

The only gotcha is that the description texts for "Local vs. global" and "About us" need to be roughly the same length so that they have the same number of lines on the page (both having three lines, both having four lines, etc):

screen shot 2019-02-07 at 12 20 48

If one has three lines and the other four, the boxes end up as different heights which doesn't look right. If there's no way to get them to be the same length, use whatever translations you think are best and send the pull request back to me; I'll try to adjust the styles a bit more to work around it.

Thanks!


marliekeverweij:
I have updated the translations. The Dutch translations for the boxes also takes three lines, so that worked out fine:
screen shot 2019-02-07 at 14 32 40

I have used the one-liner that @michieldenhaan posted on Slack and added an exclamation mark to give it some more energy! We can always update it when someone has a better one 🙂

PR #3038, opened 2019-05-23
Saving and resuming MYC sessions

antw:
(Automatically) saves MYC sessions so that a user may revisit it later.

There are a couple of new translations required in [config/locales/nl_multi_year_charts.yml](https://github.com/quintel/etmodel/blob/myc-persistence/config/locales/nl_multi_year_charts.yml).

It looks like this:

Screen Shot 2019-05-23 at 17 35 05


redekok:
> There are a couple of new translations required in config/locales/nl_multi_year_charts.yml.

Done!

antw:
Thanks for adding the translations @redekok!

Issue #3033, opened 2019-05-13
Dashboard-update: change order of items and remove bio-footprint 


marliekeverweij:
It would be nice to change the order of the dashboard items. This is something that has to be done in the database and cannot be done in the admin-section unfortunately. The preferred order would be: 
1. co2_reduction
2. energy_use
3. total_energy_cost
4. renewable_percentage
5. net_energy_import
6. blackout_hours
7. biomass_primary_demand

Also, we would like remove the bio-footprint item.

Would be nice if this could be fixed 🙂

michieldenhaan:
Related (?) to this: When I change the CO2 dashboard item to 'CO2 relative to start year' the order of the dashboard items changes:

![Energy Transition Model - Your free, independent, comprehensive, fact-based scenario builder  2019-05-21 13-26-26(1)](https://user-images.githubusercontent.com/32056448/58092393-44d70d80-7bcc-11e9-836a-a7b0fcb1db87.png)

![Energy Transition Model - Your free, independent, comprehensive, fact-based scenario builder  2019-05-21 13-26-37(1)](https://user-images.githubusercontent.com/32056448/58092402-49032b00-7bcc-11e9-8ea6-d0482f662f78.png)

redekok:
This has been changed in the ETModel database that we use for the current biomass project. When the new ETModel database will be adopted on BETA, the order should be updated!

PR #3046, opened 2019-06-05
Fix translations and give PJ unit to new biomass items

marliekeverweij:
I have given units to the new biomass dashboard items on BETA. However, now the unit (`PJ`) is fixed. Would be nice if this unit would automatically scale! Do you maybe have time to fix this @antw ?

For example now for a small dataset the dashboard item shows:
Screen Shot 2019-06-05 at 16 07 13

Related issue: https://github.com/quintel/etmodel/issues/3033


antw:
> Do you maybe have time to fix this @antw ?

It's done! PR looks good to me. :+1:

Screen Shot 2019-06-06 at 13 26 25

Issue #3036, opened 2019-05-15
Add an English and Dutch version of the workflow-image to the ETM landing page


marliekeverweij:
1. First of all it should be possible to load a static image in the chart holder instead of a chart. I first thought we did this before with the import/export images (which are gone now! see: https://github.com/quintel/documentation/blob/master/general/import_export.md), but these actually were static charts and not images, right?

2. Then the next step would be to switch between two versions of the workflow (depending on the chosen language): a Dutch and English version of the workflow.

How much work would it be to realise the above two points, @antw ? And how could we do it?

FYI: https://3.basecamp.com/3797409/buckets/11228015/messages/1740072737)

antw:
Since we don't have a way to load images instead of charts _at all_, I don't foresee any problems coding it so that the correct one is loaded depending on the currently selected language. We would just have two versions of the image, for example:

* introduction.en.png
* introduction.nl.png

antw:
I've realised that this is much easier than I thought, since we can (ab)use the system we already have in place for creating HTML tables (merit order table, plant profitability, etc).

@marliekeverweij I can set up the workflow image for you – adding the output element, creating the template, etc – if you can provide the finished images when they're ready. The only constraint is that they should be 600 pixels wide in order to fit into the space where we place charts; any larger or smaller will cause them to appear slightly blurred. *If possible*, it would be nice to also have 1200px wide versions for high-resolution screens (such as on the Macbook).

redekok:
> if you can provide the finished images when they're ready

I've just sent you the link to the images!

antw:
> I've just sent you the link to the images!

Thanks! This is what it looks like:

![Screen Shot 2019-06-12 at 13 14 36](https://user-images.githubusercontent.com/4383/59350524-3768fb00-8d14-11e9-8e41-c235ce4a2c4e.png)

I like the idea of the image a lot, but I'm not 100% sold on how it looks in the page. I have a couple of minor comments:

* Could we crop the image around the edges of the arrows/boxes and remove the extra transparent space?

* I know the dark blue is the same colour as the header text, but it feels a bit out of place to me. Would it be possible to try using the same blue as the dashboard "Change" button (`#70a2cf`)? This might be *too* blue.

* Perhaps we could also try setting the grey boxes to be the very light blue used when a chart is loading (`#f5f8fb`). This also might be too much blue, and it might be too light.

[...]

PR #3050, opened 2019-06-11
Added the default Tab to the seed data so a fresh setup can start a n…

sfp-kemp:
Added the default Tab to the seed data so a fresh setup can start a n…ew scenario


antw:
The seed file is pretty out-of-date, having not been refreshed in over a year. There are a good number of new slides, sliders, chart changes since it was made; I'm guessing if you run ETModel with the data it creates, sooner or later there will be an slider that doesn't work (because it no longer exists) or a chart that won't load because the queries have changed.

I can generate a new copy based on beta... it's pretty noisy because `rake db:seed:dump` has changed how it formats the data in the file which causes the whole thing to be rewritten (4720 insertions, 2270 deletions).

---

@sfp-kemp asked on Slack whether we really need this much data to run the app. If you want to run a local copy that mirrors what we have on beta, the answer is "yes". If you want to develop a new feature that doesn't involve charts/sliders, then probably not.

However, I think including this data is part of out commitment to being open-source; it doesn't make much sense to me for our research data (ETSource), the calculation (ETEngine, Merit, etc) and front-end (ETModel) to be open, but then not to provide a copy of the data needed to actually run ETModel. I think we should continue to bundle this data, but perhaps we should add it to our production deploy checklist to remember to refresh it regularly.

Alternatively we can change the seed file to contain only the bare minimum (I'm not sure what this constitutes), but I still think we should provide a full seed file _somewhere_ as a download, or a GitHub gist.

antw:
> but perhaps we should add it to our production deploy checklist to remember to refresh it regularly.

This gave me a thought: perhaps it *would* be better to strip seeds.rb to the bare essentials needed to run the app – I'm guessing the tabs, a slide, slider, and one output element (the upcoming [workflow image](https://github.com/quintel/etmodel/issues/3036) would suffice). With each *production* deploy we could make a habit of generating a seed.rb file based on the production (not beta) data, and provide it with [the GitHub release](https://github.com/quintel/etmodel/releases/tag/production-2018-06-20) as a downloadable asset.

This would allow third-parties to set up a complete copy of ETModel, but removes the expectation that the seed file is always up-to-date. QI staff have better ways to get copies of the beta data, so this shouldn't affect *our* workflow.

PR #3054, opened 2019-06-17
added saved_scenarios as a resource to enable reopening them

sfp-kemp:


sfp-kemp:
I broke something during squashing commits.

PR #3057, opened 2019-06-17
added saved_scenarios as a resource to enable reopening

sfp-kemp:


app/services/update_saved_scenario.rb
```diff
@@ -0,0 +1,62 @@
...
+  def failure
+    ServiceResult.failure saved_scenario.errors.full_messages
+  end
+
+  def api_scenario
+    api_response.value
+  end
+
+  def api_response
+    @_resp ||= CreateAPIScenario.call settings.merge(scenario_id: scenario_id)
```
antw: I think the scenario title needs to be included here, otherwise it defaults to "API". This means that when you use the "Save scenario" button, the scenario appears to be renamed:

Screen Shot 2019-06-18 at 12 07 37

db/migrate/20190613095331_add_scenario_id_history_to_saved_scenarios.rb
```diff
@@ -0,0 +1,5 @@
+class AddScenarioIdHistoryToSavedScenarios < ActiveRecord::Migration[5.2]
+  def change
+    add_column :saved_scenarios, :scenario_id_history, :string
```
antw: This is super-nitpicky, but could we add `after: :scenario_id` here so that the column is created after the existing `scenario_id` column instead of `updated_at`?

spec/services/update_saved_scenario_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    describe '#value' do
+      subject { result.value }
+      it {is_expected.to be_a SavedScenario}
+      it {is_expected.to be_persisted}
+    end
+
+    it 'changes the scenario_id on the SavedScenario' do
+      expect{ result }
+        .to change{saved_scenario.scenario_id}
```
antw: Apologies for nit-picking over code-style, but could we have a space before and inside blocks? (ditto for a couple of other places in the specs)

e.g. `change { saved_scenario.scenario_id }`

sfp-kemp:
@antw I think I solved all the nit-picks!

app/models/saved_scenario.rb
```diff
@@ -30,17 +32,33 @@ def self.batch_load(saved_scenarios)
...
+      else
+        @scenario ||= Api::Scenario.find(scenario_id)
+      end
     rescue ActiveResource::ResourceNotFound
       nil
     end
   end
 
+  def add_id_to_history(scenario_id)
+    scenario_id_history.shift if scenario_id_history.count <= 20
```
antw: I think this was meant to be `>= 20`?

sfp-kemp: I cut a corner while testing.. sorry!

antw:
Thanks @sfp-kemp! It seems to work really well. I've merged this with `master` manually in order to fix a merge conflict. It'll be on the beta server momentarily.

Thanks again!

Issue #3039, opened 2019-05-24
We should move the 'save scenario'-button to a more prominent place!


marliekeverweij:
@MartLubben mentioned this week that one time during a session he accidentally clicked on 'reset scenario' instead of 'save scenario', which is kind of unpractical when you just finished a scenario....

![Screen Shot 2019-05-24 at 16 53 59](https://user-images.githubusercontent.com/32833996/58337017-2760a880-7e45-11e9-984d-95a4c335cb84.png)

Thinking about this; as we want to give more attention to saving a scenario, we should mabye move the 'save scenario' button to another place!

My preference would go to underneath the Results-tab! In line with the our workflow 😎
@AlexanderWirtz , @ChaelKruip , @redekok , @DorinevanderVlies @michieldenhaan @MartLubben @michieldenhaan

MartLubben:
>is kind of unpractical when you just finished a scenario....
I couldn't agree more. 

Fortunately, is was a very easy scenario which I could reproduce :p.

redekok:
I agree to emphasize the "save" functionality! 

> My preference would go to underneath the Results-tab!

As a new tab? Or as a sidebar item under Results?

marliekeverweij:
> As a new tab? Or as a sidebar item under Results?

As a new tab. Then the tabs exactly correspond with our workflow!

![Screen Shot 2019-05-27 at 09 23 18](https://user-images.githubusercontent.com/32833996/58402845-83a01400-8061-11e9-8bc0-b6d8806c56d9.png)

marliekeverweij:
Oh I have a better idea: the new tab 'Save scenario' does not directly have to link you to the 'save scenario' page. We could also do it like this: write a good text for users about how to save a scenario and how to share your scenario with others and create a button that links to the 'save scenario' page. I think it's good for users to know what the impact of the 'save scenario'-button is **before they leave the scenario**. What do you think?

![Screen Shot 2019-05-27 at 09 23 18](https://user-images.githubusercontent.com/32833996/58462697-ab12e180-8132-11e9-8334-876daa8df794.png)

MartLubben:
>What do you think?

Very good idea!!

marliekeverweij:
I think that it is achievable to create this new button before the deploy on July 10th, therefore I have added the deploy-milestone to this issue. Please let me know when I am wrong!

marliekeverweij:
[...]

Issue #2123, opened 2016-05-24
User should have option to 'reset' to saved values in stead of default values


dennisquintel:
### Problem

I find it quite annoying, when you are working in a non-default scenario, often you want to go back to your 'original' slider setting, but all you can do is reset to the default setting. With singular sliders: that is doable to write down or remember the slider setting, but when you are changing a slider group, all you can do is reload the entire scenario. 😒 
#### Slider reset

![image](https://cloud.githubusercontent.com/assets/104830/15499373/2b439c74-21a4-11e6-88f5-f24fddfc2ead.png)
### Possible solution

We might want to add a select box in the menu where you can choose your behaviour.

![image](https://cloud.githubusercontent.com/assets/104830/15495460/60897746-2190-11e6-9200-215d1cf3df35.png)

cc @grdw @ChaelKruip @AlexanderWirtz

grdw:
There are multiple ways of achieving this I guess. One way would be store all slider settings in the browser's `LocalStorage` (if the customer in question does not use [IE7, IE6 or Opera Mini](http://caniuse.com/#search=localstorage)). Than I think you could easily reset to `Last saved`. Another more complex way would be storing multiple scenario's (like a simple versioning system). 

I like the general idea though. 👍

dennisschoenmakers:
Good idea. We could also use the preset_scenario's `user_values`.

dennisschoenmakers:
What do you think @antw?

antw:
> We could also use the preset_scenario's user_values.

Agree. I assume that "Last saved" refers to the most recent time the scenario was explicitly saved in ETModel. That being the case, we should likely set the `preset_scenario_id` of the *active* scenario to point at the saved scenario, and use those values when resetting.

The question is whether "Last saved" needs to be a user-selectable option at all; I think it should just be the default behaviour.

dennisschoenmakers:
I realize now,  I could have been clearer in my description: with 'resetting' I mean hitting this button:

![image](https://cloud.githubusercontent.com/assets/104830/15499345/05fbc478-21a4-11e6-90ce-a1e47b4ae369.png)

I am adjusting the body of this ticket accordingly.

antw:
Okay! So reset the slider to the value stored in the preset/saved scenario, or if no value is present there, back to the default. A big :+1: from me.

[...]

PR #3069, opened 2019-07-02
Ask user to confirm before resetting their scenario

antw:
Show a pop-up message asking the user to confirm that they wish to reset their scenario before doing so. Intended to prevent accidental mis-clicks. :wink:

Would you mind adding an appropriate translation in [config/locales/nl_etm.yml](https://github.com/quintel/etmodel/blob/41f7434f0dffdb43ad6b34413a0fe0e559850dbb/config/locales/nl_etm.yml#L63)?

Please reassign as you see fit.

---

Ref #3039

### DIFF
--- a/app/views/layouts/etm/_settings_menu.html.haml
+++ b/app/views/layouts/etm/_settings_menu.html.haml
@@ -32,7 +32,7 @@
 
         %a.save_as{:href => new_scenario_path}= t("header.save_scenario_as")
 
-        %a#reset_scenario{:href => scenario_reset_path}= t("header.reset_scenario")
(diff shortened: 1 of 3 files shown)


marliekeverweij:
Nice that you have implemented this!! I've added the Dutch translation:
Screen Shot 2019-07-03 at 10 37 41

A question out of curiosity:
Does this button reset the scenario to the initial start values for that dataset?
OR
Does this button reset the scenario to the latest saved slider settings?

marliekeverweij:
> A question out of curiosity:
> Does this button reset the scenario to the initial start values for that dataset?
> OR
> Does this button reset the scenario to the latest saved slider settings?

I think I can answer my own question: I just saw [this issue].(https://github.com/quintel/etmodel/issues/2123)

**The "Reset scenario"-button brings you back to the initial start values for that dataset**

PR #3099, opened 2019-07-10
Whats new

MartLubben:


antw:
> Not sure about the hound things tho.

It can be safely ignored: it's a code-style checker, but doesn't mean that anything is wrong. I might tweak these files a bit since they're mixing HTML and HAML syntax, but don't let that stop you from merging.

Issue #3107, opened 2019-07-18
Add link to factsheet to "show saved_scenario" page


sfp-kemp:
At the moment user's cant create factsheets from saved_scenarios.

the pages with the following path: [/saved_scenarios/:id](https://beta-pro.energytransitionmodel.com/saved_scenarios/4369) should provide a link to https://beta-pro.energytransitionmodel.com/scenarios/:id.

michieldenhaan:
Closing, fixed in https://github.com/quintel/etmodel/pull/3108

PR #3108, opened 2019-07-19
Added load factsheet button to show saved scenario page

sfp-kemp:
Some users want to share factsheets about scenarios that are stable over for a long time. 
At the moment there is no easy way to generate a factsheet from a saved scenario. The current path always leads the user to an "API scenario".

This change enables a user to load a factsheet directly from the "show saved_scenario page".  The reason that I opted for this solution over the one in #3107 is that it has less impact in both development time and maintainability and it is closer to the workflow previously employed by users.

### DIFF
--- a/app/views/saved_scenarios/show.html.haml
+++ b/app/views/saved_scenarios/show.html.haml
@@ -12,7 +12,7 @@
 
   = link_to t("scenario.load_factsheet"),
             factsheet_scenario_path(@saved_scenario.scenario_id),
-            class: "button primary"
+            class: "button"
 
   - if @scenario.description
     %em= @description
--- a/app/views/saved_scenarios/show.html.haml
+++ b/app/views/saved_scenarios/show.html.haml
@@ -4,10 +4,18 @@
     .warning
       = @warning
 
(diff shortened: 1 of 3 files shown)


app/views/saved_scenarios/show.html.haml
```diff
@@ -4,12 +4,16 @@
...
+  %h1= (@scenario.title or t 'scenario.no_title' ).truncate(40, separator: ' ')
 
-  = link_to t("scenario.load"), load_saved_scenario_path(@saved_scenario), class: "button primary"
+  = link_to t("scenario.load"),
+            load_saved_scenario_path(@saved_scenario),
+            class: "button primary"
+
+  = link_to t("scenario.load_factsheet"),
+            factsheet_scenario_path(@saved_scenario.scenario_id),
+            class: "button primary"
```
antw: Could we perhaps remove the `primary` class from "Open infographic" so that the "Open this scenario" button stands out more?

sfp-kemp: good idea

PR #3112, opened 2019-08-14
restyled load scenario page

sfp-kemp:
While working on combining scenarios I've cleaned up this view a bit. We might want to merge it. What do you think @antw ?

### DIFF
--- a/app/views/saved_scenarios/show.html.haml
+++ b/app/views/saved_scenarios/show.html.haml
@@ -1,34 +1,48 @@
-.scenario_wrapper#show_scenario
-
+%main.saved-scenario__main
   - if @warning
     .warning
       = @warning
 
-  %h1= (@scenario.title or t 'scenario.no_title' ).truncate(40, separator: ' ')
+  .saved-scenario__title
+    %h1= (@scenario.title or t 'scenario.no_title' ).capitalize
 
-  = link_to t("scenario.load"),
-            load_saved_scenario_path(@saved_scenario),
-            class: "button primary"
+  .saved-scenario__content
+    .saved-scenario__description
+      .saved-scenario__subhead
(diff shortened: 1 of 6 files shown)


app/views/saved_scenarios/show.html.haml
```diff
@@ -1,34 +1,52 @@
...
-      %strong #{t('scenario.area_code')}:
-      = t @scenario.area_code
-    %li
-      %strong #{t('scenario.end_year')}:
-      = @scenario.end_year
-
-  %p
-    %small{:style => "font-size: 12px; font-weight: normal"}
-      = t("scenario.overwritten")
+      .saved-scenario__field
```
antw: Since these are basically "label" => "value", I think [a definition list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) would be a better fit than `div`s and `p`aragraphs.

app/views/saved_scenarios/show.html.haml
```diff
@@ -1,34 +1,52 @@
...
+                  title: t('saved_scenario.tooltips.open')
+
+        = link_to "", factsheet_scenario_path(@saved_scenario.scenario_id),
+                  class: "saved-scenario__nav tooltip fa fa-bar-chart",
+                  title: t('saved_scenario.tooltips.factsheet')
 
-  = link_to t("scenario.load"),
-            load_saved_scenario_path(@saved_scenario),
-            class: "button primary"
+        -# = link_to "",
```
antw: I think this commented-out "transport" link can be removed?

sfp-kemp:
I'm having a bit of trouble formulating the issues I'm trying to solve here. I'll get back on it when I've structured my thoughts a bit better

PR #3113, opened 2019-08-16
Update DB and tables to use utf8mb4

antw:
Updates the database and tables to use `utf8mb4` character set and `utf8mb4_unicode_ci` collation. Hopefully this will fix the constantly changing schema.rb. 😅 

Changing the character set seems to bump TEXT columns to MEDIUMTEXT (perhaps because the extra bytes might not fit in TEXT?). In most cases I don't think this extra space is needed so I adjusted those columns back down to 64kb.

### DIFF
--- a/db/migrate/20190816120851_set_mysql_encoding.rb
+++ b/db/migrate/20190816120851_set_mysql_encoding.rb
@@ -1,0 +1,64 @@
+class SetMysqlEncoding < ActiveRecord::Migration[5.2]
+  def up
+    unless ActiveRecord::Base.connection.adapter_name.downcase =~ /^mysql/
+      return
+    end
+
+    ActiveRecord::Base.transaction do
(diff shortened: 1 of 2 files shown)


antw:
I'll look into the Semaphore failure. I think it has to do with indexed VARCHAR columns being too long. Odd that it works locally; maybe different versions of MySQL?

sfp-kemp:
It works on my machine as well.... 
The error happens in db:setup. I haven't run that command yet and am not to keen on running it right now.

antw:
I remember this from a similar migration in ETEngine. It's [a limitation (?) of MySQL 5.6 and earlier](https://stackoverflow.com/a/1814594), which Semaphore uses. Some development machines may also be running 5.6; I've adjusted the migration to reduce the max length of the affected columns to 191, which I think should be plenty.

I'll see what Semaphore says. 🤞

PR #3120, opened 2019-08-27
Biomass

marliekeverweij:
Adding all ETModel changes related to 'the biomass project'. The most important addition is the new sankey for biomass! 
image
Furthermore this PR includes:
- Translations of new charts
- Tranlsations of new slider


antw:
> Could you generate a new schema.rb from the merged ETModel database?

It looks like the database used for biomass has a couple of old ETEngine tables in there. I expect someone accidentally imported an ETEngine dump into their local ETModel DB using Sequel Pro. This can be fixed by getting a fresh DB dump from dbbot and importing it with `bin/import-db` (_not_ Sequel Pro, which will preserve the ETEngine tables).

I've fixed schema.rb by removing the extra tables.

marliekeverweij:
Okay good to know! That could be me, I do not update the ETEngine database often.

PR #3124, opened 2019-09-05
Embedded PICO into the dutch supply -> wind slide

sfp-kemp:


spec/models/embeds/pico_area_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+
+describe Embeds::PicoArea, vcr: true do
```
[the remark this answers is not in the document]
sfp-kemp: 😨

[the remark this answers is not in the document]
antw: Ignore that, it seems silly for that metric to be enabled in RSpec blocks.

app/assets/javascripts/embeds.js
```diff
@@ -0,0 +1,81 @@
...
+  * {
+  *   action: "updateInlandWindTurbine",
+  *   argument: {
+  *     power: pico.inMegaWatt(power)
+  *   }
+  * }
+
+*/
+
+embeds = {
```
antw: Should this perhaps be `var embeds`?

app/assets/javascripts/embeds.js
```diff
@@ -0,0 +1,81 @@
...
+  attach: function(){
+    window.addEventListener('message', embeds.receiveMessage, false)
+  },
+
+  // dependencies:
+  //   - jquery
+  //   - fancybox
+  close: function(event){
+    $.fancybox.close()
```
antw: I think this needs to also remove the "message" event that was added in `attach`.

```suggestion
    $.fancybox.close()
    window.removeEventListener('message', embeds.receiveMessage)
```

app/assets/javascripts/embeds.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+  },
+
+  // Helper functions
+  availableActions: function(){
+    return Object.getOwnPropertyNames(embeds.actions)
+  },
+
+  actionIsAvailable: function(actionName){
+    return embeds.availableActions().includes(actionName)
```
antw: [IE11 doesn't support `includes` unfortunately](https://caniuse.com/#search=includes).
```suggestion
    return embeds.availableActions().indexOf(actionName) !== -1
```

app/controllers/embeds/picos_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+        set_scenario_area
+        @pico_module = params[:pico_module]
+      end
+    end
+  end
+
+  private
+
+  def set_scenario_area
+    scenario = Api::Scenario.find(Current.setting.api_session_id)
```
antw: I'm not sure if this is necessary. If you only want the area code, it should be available as `Current.setting.area_code` and will save the extra request to the API.

app/models/embeds/pico_area.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def type
+    pattern_whitelist.each do |type, pattern|
+      return type if area.match?(pattern)
+    end
+    return :land
+  end
+
+  def to_json
```
antw: If this is send as a request response, you can instead define `as_json` returning a Ruby hash, and Rails will take care of the conversion and escaping the type and name.

```ruby
def as_json(*)
  { area_type: type, area_name: area_name }
end
```

app/assets/javascripts/embeds.js
sfp-kemp: > Look mostly good to me, Hound's nagging aside. 😆
> 
> Two minor comments:
> 
>     * Clicking the PICO link in an unsupported region (e.g. France) brings up PICO for the Netherlands. Could we instead check that the area is supported, and if not display a message informing the user?
good idea. I'll add it after I've taken care of 🐶 

>     * When ready to merge, would you please squash this down to remove the "work in progress" commits? e.g. • Update Webmock for better Ruby 2.6 support  • Added first version of PICO integration  • Any other commits which are relevant as separate commits

I was planning on squashing it all down to a single commit later on. These commits are for review purposes!

app/models/embeds/pico_area.rb
sfp-kemp: This is a bit of a hack. I'm using it to inject a javascript object literal into an onclick event handler(in embeds/picos/show.html). Its not meant to be used in many places. 

Since the class is just a decorator I didn't have problems with doing this. Are you of a different opinion?

app/controllers/embeds/picos_controller.rb
sfp-kemp: I thought i tried that and failed. I might have done something wrong though. I'll retry.

app/models/embeds/pico_area.rb
sfp-kemp: @antw I'd love to hear your opinion about it before I merge!

antw: > Are you of a different opinion?

It's not a strong opinion, no. :smiley: The only concern I'd have is if the area name has a `'` in it, the JSON will be invalid.

antw: > The only concern I'd have is if the area name has a ' in it, the JSON will be invalid.

... now that I write that, I'm realising that `to_json` is invalid JSON; the keys and values should both be surrounded with double quotes.

app/models/embeds/pico_area.rb
```diff
@@ -0,0 +1,41 @@
...
+        return type if area.match?(pattern)
+      end
+      :land
+    end
+
+    def to_json(*_args)
+      "{areaType:'#{type}', areaName:'#{area_name}'}"
+    end
+
+    def is_supported?
```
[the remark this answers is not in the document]
sfp-kemp: g

Issue #3089, opened 2019-07-09
Local vs. Global: translations of regions are somehow not shown


marliekeverweij:
When I pick a region other than the Netherlands, the translation is somehow not shown...
![Screen Shot 2019-07-09 at 09 28 18](https://user-images.githubusercontent.com/32833996/60867891-1f5d9c00-a22c-11e9-9227-bb4f28a0910c.png)

antw:
This is because area name translations aren't available to the JavaScript front-end code. It can be fixed, but not without fallout everywhere that we display area names in ETModel. I'm not confident we'll be able to make this change and find all the places we use area names in time for this milestone. I think we'll have to tackle this after the deploy.

ChaelKruip:
> This is because area name translations aren't available to the JavaScript front-end code.

Including @sfp-kemp as he has some ideas how to improve maintenance of translations and other front-end items (side-bar items etc.).

marliekeverweij:
@antw, how much work would it approximately be to fix this (minor) issue?

PR #3128, opened 2019-09-23
Add area name translations to Local vs. Global

antw:
Moves area translations to a namespace so that they may be included in the JavaScript translations used by Local vs. Global.

This affects everywhere in the ETM where we show translated area names. I think I got them all, but I'd be grateful for a second pair of eye checking that everything looks correct.

Ref #3089


marliekeverweij:
I've checked the:
- CO2 factsheet
- Energymix factsheet
- ETM in general
- Local vs. Global
- Scenario report

I only found missing translations in the scenario report:
Screen Shot 2019-09-24 at 10 04 24
Screen Shot 2019-09-24 at 10 04 08

I couldn't open the multi-year charts,  somehow it was not visible:
Screen Shot 2019-09-24 at 10 06 08

antw:
I fixed the missing translation in the report title. The "1.1xAREA_CODE" is something we do with national scenarios (e.g. "1.1xNL"); we don't usually show the area code for smaller datasets because it doesn't look nice. I've changed the regional report to show just "1.1x", which is consistent with the dashboard.

I've checked the multi-year charts pages myself, so they'll be fine. They only show up in your local copy [if your config/config.yml has a `multi_year_charts_url` specified](https://github.com/quintel/etmodel/blob/89af47b6e6db00c6d3e61a80662d174b7d217f20/config/config.sample.yml#L11-L12).

PR #3140, opened 2019-10-03
Update "what's new" pages for the upcoming deploy

redekok:
@marliekeverweij @ChaelKruip Do you know how to disable the message below until the deploy is done?

image


antw:
A few minor comments for next time:

* Be sure to merge with `master` rather than `production`. The production branch gets overwritten during the deploy, so anything that isn't on master will be lost.

* You can delete the old images in app/assets/images/pages/whats_new, since we don't need them after the page has been updated.

* Images can be compressed with [ImageOptim](https://imageoptim.com/mac) to reduce sizes by 30-40%. I've done this for the new "what's new" images:

  Screenshot 2019-10-08 at 14 27 05

> Do you know how to disable the message below until the deploy is done?

Where was the screenshot taken? The "what's new" banner should not have shown up on production – it is controlled by [the `whats_new_cutoff` setting in config/config.yml](https://github.com/quintel/etmodel/blob/2200afaa316a32e801f18900045149911c0f6048/config/config.sample.yml#L53-L56). If it's on your local machine, it will default to always be shown; you can hide the banner by adding that option to your config/config.yml with a date in the past.

ChaelKruip:
> You can delete the old images in app/assets/images/pages/whats_new, since we don't need them after the page has been updated.

I think it would be nice to be able to show something like a change log. Would it be possible to keep the old pages and let the user click back to previous deploys? In other words: I have the feeling that we should not delete old stuff as it might be useful...

antw:
> I have the feeling that we should not delete old stuff as it might be useful...

[They'll still exist in the Git history](https://github.com/quintel/etmodel/tree/0274e45bed0b9c191c990a36b6fbcf5abbd02762/app/assets/images/pages/whats_new). It's easy enough to recover them.

redekok:
Thanks for your comments, @antw! 

> Be sure to merge with master rather than production. The production branch gets overwritten during the deploy, so anything that isn't on master will be lost.

That's inconvenient, indeed. Next time I'll first merge the changes into master and then cherry-pick the relevant commits to pro (since we do like to share the https://pro.energytransitionmodel.com/whats-new link to our power users prior to the deploy).  

> Where was the screenshot taken?

It was indeed taken on my local machine. I was just making sure the banner wouldn't pop up on the LIVE server yet before the deploy. :)

PR #3158, opened 2019-11-08
added csv form support to saved scenarios

sfp-kemp:
This was requested by @AlexanderWirtz but both @jorisberkhout and @MartLubben want this feature. The change isn't thought out that well but its minor and doesn't advertise itself within the UI. I think its a safe addition. Can we merge this?


antw:
@AlexanderWirtz @jorisberkhout @MartLubben: The CSV file, as written here, would include a row for all of the ETModel inputs, even those for which the user has not set a value. However, for those without a user value, the value is left blank. Is that what you're looking for?

For example:

| | |
| :-- | --: |
| input_one | |
| input_two | |
| input_three | 1.0 |
| input_four | 3.0 |
| input_five | |

Would you prefer it to omit rows which don't have a value?

AlexanderWirtz:
> The CSV file, as written here, would include a row for all of the ETModel inputs, even those for which the user has not set a value. However, for those without a user value, the value is left blank. Is that what you're looking for?

Yes, @antw that was the idea. This is useful to some users, as it means that they can easily import it into an Excel file and the order will remain the same regardless of how many sliders were changed in a scenario. We've had requests from both Gasunie and TenneT to implement this feature, as this allows them to quickly compare scenarios in terms of inputs, as well as outputs (which have been downloadable for a while now). Once this is on the Live server I will notify TenneT.  

As @sfp-kemp mentioned, and we have discussed before, it would also be great if users can change said input value list and re-upload it to create a scenario in one go. I remember from past discussions that this is not trivial, though.

PR #3172, opened 2019-11-18
Co2 sheets toggle other emissions

noracato:
Add an extra link/toggle in the footer of the chart by which you can turn off the "other, non energetic, greenhouse emissions". The chart defaults to showing these emissions.

I refactored the charts.js to have more influence on which queries are added to the charts. I'm not sure if it's very efficient as most queries are now in vars. 

There are also some extra gqueries needed for the total emissions - I will create an etsource pull request as well, called co2_sheets.


app/assets/javascripts/establishment_shot/charts.js
```diff
@@ -103,62 +139,55 @@ EstablishmentShot.Charts = (function () {
...
         getQueries: function () {
             var chart,
-                queries = [];
+                queries = [],
+                charts = this.getCharts();
 
-            for (chart in this.charts) {
-                this.charts[chart].series.forEach(function (serie) {
+            for (chart in charts) {
+                charts[chart].series.forEach(function (serie) {
```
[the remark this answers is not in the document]
antw: I think this can be ignored... it seems like a false-positive. I think it's warning about [this kind of error](https://stackoverflow.com/a/46027349), which doesn't apply in this case as far as I can tell. What do you think @sfp-kemp?

app/assets/javascripts/establishment_shot/charts.js
```diff
@@ -26,19 +23,70 @@ EstablishmentShot.Charts = (function () {
...
+                        { key: 'co2_sheet_industry_energy_total_co2_emissions_only_energetic',
+                          title: 'co2_sheet_industry_energy_total_co2_emissions',
+                          fa_icon: '\uf275' },
+                        { key: 'co2_sheet_transport_total_co2_emissions_only_energetic',
+                          title: 'co2_sheet_transport_total_co2_emissions',
+                          fa_icon: '\uf1b9' },
+                        { key: 'co2_sheet_buildings_households_total_co2_emissions_only_energetic',
+                          title: 'co2_sheet_buildings_households_total_co2_emissions',
+                          fa_icon: '\uf015' }
+                    ],
```
antw: No biggie, but the indenting is a bit off here.

noracato: Thanks! I fixed the indent issue and will merge the PR as one commit when all checks have completed.

PR #3178, opened 2019-11-26
Picofix

sfp-kemp:


app/models/embeds/pico/area_type.rb
```diff
@@ -0,0 +1,64 @@
...
+      #
+      # matcher:               A regex to detect what area type an Api::Area is
+      #                        from its #area.
+      #
+      # select_field_stripper: Strategy for creating a select_field from
+      #                        an Api::Area#area.
+
+      type =
+        Struct.new(:key, :select_field, :matcher, :select_field_stripper) do
+          def get_select_value(area)
```
antw: Very nitpicky, but `get_` doesn't feel very Ruby-like to me. How would you feel about calling this `select_value`?

app/models/embeds/pico_area.rb
```diff
@@ -3,39 +3,38 @@
...
     def type
-      pattern_whitelist.each do |type, pattern|
-        return type if area.match?(pattern)
-      end
-      :land
+      TYPES.find { |type| area.match?(type.matcher) } || FALLBACK_TYPE
+    end
+
+    def select_value
+      type.get_select_value(area)
```
antw: If the area type is a `Res`, an `UnmatchableSelectValueError` is raised when calling `get_select_value`. Perhaps we should show a message informing the user that the current region isn't supported?

app/models/embeds/pico/area_type.rb
sfp-kemp: I don't really agree with this. If it was an accessor it would. I feel get implies an argument. Maybe 'select_value_from' would be a bit more clear?

antw: >  I feel get_select_value implies an argument.

Fair point. :+1: `select_value_from` sounds good.

sfp-kemp:
@antw Both issues have been resolved. Can I squash and merge this?

PR #3182, opened 2019-12-05
Add Overmorgen CSV dump

noracato:
As requested by @MartLubben. Create a csv report based on a yml file containing queries (should be in config/saved_scenario_reports). The report can be found only through url, because it's only valid for Over Morgen. 
In the future, more types of reports based on different queries can be created by simply adding a yml file to the config/saved_scenario_reports folder.


app/controllers/saved_scenario_reports_controller.rb
```diff
@@ -0,0 +1,52 @@
+# Create a report based on a .yml file in config/saved_scenario_reports
+# The report name and format should be provided in params (only works for csv)
+class SavedScenarioReportsController < ApplicationController
+  before_action :ensure_valid_browser
+
+  def show
```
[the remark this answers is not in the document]
noracato: How can I make the method shorter?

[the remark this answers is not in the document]
antw: > How can I make the method shorter?

I'm not sure you can. The simplest way to reduce the ABC metric is to split the method into separate methods. For example, you might have `render_success` and `render_failure`...

```ruby
  def show
    @scenario_id = SavedScenario.find(params[:saved_scenario_id]).scenario_id
    @yml = YAML.load_file("config/saved_scenario_reports/#{report_name}.yml")

    if valid_api_response? && valid_report_name?
      render_success(api_response['gqueries'])
    else
      render_failure
    end
  end
```

... however I'm not sure that really makes the method any simpler to read. Personally I think it's fine as it is, and I'd be inclined to ignore the warning.

app/controllers/saved_scenario_reports_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    @scenario_id = SavedScenario.find(params[:saved_scenario_id]).scenario_id
+    @yml = YAML.load_file("config/saved_scenario_reports/#{report_name}.yml")
+
+    if valid_api_response? && valid_report_name?
+      @query_api_response = api_response['gqueries']
+      respond_to do |format|
+        format.csv { render 'saved_scenarios/reports/show.csv.erb' }
```
antw: The Rails convention is to put views in a directory which matches the controller name. So rather than placing the show.csv.erb file in app/views/saved_scenarios/reports, it would instead go in app/views/saved_scenario_reports. As a bonus, the `respond_to` block can then be simpler as Rails will find the file automatically.

```suggestion
        format.csv
```

app/views/saved_scenarios/reports/show.csv.erb
```diff
@@ -0,0 +1,8 @@
+<%=@scenario_id%>
+<%-@yml.each do |category, value| %>
+  <%-value.each do |subcategory, subvalue| %>
+    <%-subvalue.each do |split, query| %>
+      <%=category%>;<%=subcategory%>;<%=split%>;<%=@query_api_response[query]['future']%>;<%=@query_api_response[query]['unit']%>
```
antw: Being a bit nitpicky, but the CSVs that ETEngine generates use commas as separators rather than semicolons. The indenting of this line also creates extra leading whitespace in the CSV file; this might mess up some tools which don't ignore it.

```
1214450
      biomassa;;collectief;5604795553.98533;MJ
      biomassa;;individueel;21958769999.989735;MJ
      biomassa;;transport;12451999999.999996;MJ
      biomassa;;industrie;5913489241.570863;MJ
# ...
```

```suggestion
<%=category%>;<%=subcategory%>;<%=split%>;<%=@query_api_response[query]['future']%>;<%=@query_api_response[query]['unit']%>
```

Perhaps it would also be nice to add a header row (instead of the scenario ID?) to describe each column?

```csv
Category,Subcategory,Split,Value,Unit
```

app/controllers/saved_scenario_reports_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  private
+  def report_name
+    params[:report_name].split('.')[0]
+  end
+
+  def valid_api_response?
+    !api_response['errors'].present?
```
antw: Very nitpicky: this can be simplified slightly as `api_response['errors'].blank?`

`blank?` is provided by Rails and returns true if the value is false, nil, or a collection of some kind (String, Hash, Array) which is empty.

sfp-kemp:
@antw Is it ok if I merge this?

PR #3208, opened 2020-01-07
Converter source download

noracato:
* Add a button in the tech-spec table to download the source .xls file form github
* New rake task that fetches the github download urls for each converter and writes them to a .yml
* Facelift for tech-spec table


lib/tasks/fetch_converters_download_source.rake
```diff
@@ -0,0 +1,37 @@
...
+  converters.each do |converter|
+    converter = converter.gsub(/ $/, '')
+    generate_urls(converter).each do |url|
+      yml[converter] = url + '?raw=true' if valid_url?(url)
+    end
+  end
+  File.write(yml_file, yml.to_yaml)
+end
+
+def generate_urls(converter)
```
sfp-kemp: Put these methods in a module so they won't pollute the global namespace.

PR #3220, opened 2020-01-20
Remove duplicate nuclear 3rd gen and add nuclear 2rd gen

marliekeverweij:
A client pointed out this issue:

Screen Shot 2020-01-20 at 13 56 26

It is already fixed on BETA, but not on LIVE


antw:
This has been merged and deployed.

PR #3240, opened 2020-02-27
Flatten descriptions for slides and input elements

noracato:
We can now edit the descriptions of slides/sliders in the config .yml files. 

Requested by @michieldenhaan. In the current state of the YModel migration, it's still neccessary to edit the descriptions of slides and sliders through migrations. This is a pain with the deploy coming up.

I dumped these descriptions to the config files for slides and sliders and added a describable module for the YModel models.

### DIFF
--- a/app/models/input_element.rb
+++ b/app/models/input_element.rb
@@ -23,10 +23,11 @@
 # Model representing a slider in the front-end. Settings get stored in scenarios
 class InputElement < YModel::Base
   include AreaDependent::YModel
+  include Describable
 
   ENUM_UNITS = %w[radio weather-curves].freeze
 
-  has_one :description, as: :describable
+  # has_one :description, as: :describable
   belongs_to :slide
 
   class << self
@@ -95,7 +96,7 @@ class InputElement < YModel::Base
   end
 
   def has_flash_movie # rubocop:disable Naming/PredicateName
-    description.try :embeds_player?
+    description_embeds_player?
   end
 
   # For loading multiple flowplayers classname is needed instead of id
@@ -103,7 +104,7 @@ class InputElement < YModel::Base
   #
   def sanitized_description
     ie8_sanitize(
-      description.try(:sanitize_embedded_player)
+      description_sanitize_embedded_player
     ).html_safe
   end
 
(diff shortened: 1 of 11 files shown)


app/models/input_element.rb
```diff
@@ -23,10 +23,11 @@
 # Model representing a slider in the front-end. Settings get stored in scenarios
 class InputElement < YModel::Base
   include AreaDependent::YModel
+  include Describable
 
   ENUM_UNITS = %w[radio weather-curves].freeze
 
-  has_one :description, as: :describable
+  # has_one :description, as: :describable
   belongs_to :slide
```
antw:
```suggestion
  belongs_to :slide
```

spec/models/input_element_spec.rb
```diff
@@ -77,4 +77,13 @@
...
     end
   end
+
+  describe '#sanitized_description' do
+    subject do
+      ie = described_class.new(description: { 'content_en' => 'foobar' })
+      ie.sanitized_description
+    end
+
+    it { is_expected.to eq 'foobar' }
```
antw: This isn't really testing the behaviour of `sanitized_description` as there's nothing in the input string "foobar" which would be sanitized.

app/models/input_element.rb
```diff
@@ -95,15 +96,15 @@ def parsed_name_for_admin
...
+    description_embeds_player?
   end
 
   # For loading multiple flowplayers classname is needed instead of id
   # added the andand check and html_safe to clean up the helper
   #
   def sanitized_description
     ie8_sanitize(
-      description.try(:sanitize_embedded_player)
+      description_sanitize_embedded_player
```
antw: I wonder if we can drop this entirely? The only `<object>` in the input element descriptions is a YouTube video. I'm pretty certain we don't use Flowplayer anymore.

noracato: I only count 5 occurences of Flowpayer videos in all of the descriptions (and I believe those videos to be hidden either way), so I think it's safe to drop it. I could also remove those 5 video links from the descriptions, to be totally safe.

Issue #3255, opened 2020-03-02
Scenario report: incorrect scaling of curtailed electricity


DorinevanderVlies:
Scenario gives MJ as units, should be PJ

![image](https://user-images.githubusercontent.com/19907658/75670476-1ae36900-5c7d-11ea-9f2e-e3d9d7a444ef.png)

Related files/lines
- https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/vertical_stacked_bar_163_use_of_excess_electricity/electricity_curtailed.gql
- https://github.com/quintel/etmodel/blob/master/config/reports/main.en.liquid/#L132

notifying @marliekeverweij

PR #3256, opened 2020-03-02
Fix units of curtailed electricity in scenario report

DorinevanderVlies:
Fixes https://github.com/quintel/etmodel/issues/3255


marliekeverweij:
I'll add the fix for this issue (https://github.com/quintel/etsource/issues/2237#issuecomment-593351965) to the PR

PR #3268, opened 2020-03-05
Adjust slide titles for hourly curves

antw:
Adjusts the slide titles for the hourly curve downloads.

The English version is changed to slightly improve the grammar (e.g. changes "Hourly curves electricity" to "Hourly curves for electricity"). I changed the Dutch version as well, but would appreciate a second opinion on that.


marliekeverweij:
Ziet er goed uit!

PR #3278, opened 2020-03-17
Add hourly solar curtailment chart

noracato:
I believe the solar_curtailment chart is hereby finished. Thanks again @antw for helping with the chart bugs


marliekeverweij:
I love it!

I found one issue. However, I cannot reproduce it, so don't know what to do about it
Screen Shot 2020-03-18 at 09 15 42

noracato:
@antw Thanks! I incorporated your suggested changes and removed as many 'this' as possible. I agree it's much easier to read the code now.
@marliekeverweij It looks like the housholds are plotted for a week view instead of a year view. Could this have happened after switching options too fast?

marliekeverweij:
> It looks like the housholds are plotted for a week view instead of a year view. Could this have happened after switching options too fast?

Yes that could be it!

PR #3289, opened 2020-03-24
What's new page & banner deploy #2 2020

marliekeverweij:
Few comments:
- I've updated the banner text with a pre-deploy message
- I've kept the previous update messages (heat projects etc) below the new update messages and added the dates. What do you think about that @antw?
- Links to the ETM will be added after deploy
- Heat demand profiles
  - @michieldenhaan could you check the text? I do not exactly know for how many houses we've aggregated the profiels
  - we should add an image for this when the profile is definite

If people want to update texts / images, feel free to do so!

Screen Shot 2020-03-24 at 11 38 00


marliekeverweij:
Deploy date is certain now! Would be nice to put the What's new page online asap :)

PR #3292, opened 2020-03-24
Charts that only have one relatee output element show hourly toggle

### LINKED ISSUE #3290: Solar PV curtailment chart and table should be linked
In the admin-interface I tried to link the solar PV curtailment chart and table by adding a 'related (hourly) output element' to the (annual) table (number 256). This is how it usually works. However, the button does not show up in the hourly solar pv curtailment chart. Would be nice if it does!

Screen Shot 2020-03-24 at 13 52 26

noracato:
On request of @marliekeverweij. Charts that are only linked to by one yearly table now show a toggle to said table.

Closes #3290

### DIFF
--- a/app/models/output_element.rb
+++ b/app/models/output_element.rb
@@ -107,6 +107,18 @@ class OutputElement < ActiveRecord::Base
     description.present?
   end
 
+  # some charts only have one relatee output element, in that case
+  # the hourly/yearly toggle should also show on the current chart
+  def has_one_relatee_output_element?
+    relatee_output_elements.length == 1
+  end
+
+  def relatee_output_element
+    return unless has_one_relatee_output_element?
+
+    relatee_output_elements[0]
+  end
+
   # rubocop:enable Naming/PredicateName
 
   def self.select_by_group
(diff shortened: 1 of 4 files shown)


app/models/output_element.rb
```diff
@@ -107,6 +107,18 @@ def has_description?
     description.present?
   end
 
+  # some charts only have one relatee output element, in that case
+  # the hourly/yearly toggle should also show on the current chart
+  def has_one_relatee_output_element?
+    relatee_output_elements.length == 1
```
antw: One this occasion, use `count` rather than `length`. `length` will load all of the related output elements and then count them, whereas `count` will run a `SELECT COUNT` query which avoids loading data we don't care about. ActiveRecord also has a [`one?`](https://apidock.com/rails/ActiveRecord/Relation/one%3F) method.

```suggestion
    relatee_output_elements.one?
```

app/models/output_element.rb
```diff
@@ -107,6 +107,18 @@ def has_description?
...
+  end
+
+  def relatee_output_element
+    return unless has_one_relatee_output_element?
+
+    relatee_output_elements[0]
```
antw: Use `relatee_output_elements.first` if you only want the first object. Like the length/count example, ActiveRecord will be smart enough to only fetch the first output element when using `first`.


```suggestion
    relatee_output_elements.first
```

PR #3293, opened 2020-03-24
Add Mailchimp integration

antw:
Adds integration with Mailchimp to ETModel. The changes are not big. 

* When a user subscribes or unsubscribes to the newsletter through the ETM, a request is sent to Mailchimp to add or remove them from the mailing list.
* There's a hook so that if the visitor clicks the "unsubscribe" link in the e-mail itself, Mailchimp will let the ETM know. This allows us to keep the ETM in sync with the list on Mailchimp.
* Finally, if a user deletes their ETM account, we remove their data from Mailchimp entirely.

There are [a couple of texts which need translating](https://github.com/quintel/etmodel/blob/dd89608e118d5ebaf561cd9eaba875bc3b599519/config/locales/nl_users.yml#L72-L83) (sorry!). Feel free to adjust them – and the English version – as you see fit. I speculated that we'd be sending about four messages a year; please adjust that if you think otherwise.

The subscribe/unsubscribe buttons behave differently depending on the environment:

* **Production / Live**: The buttons are shown on the "My account" page; Mailchimp will be notified when a visitor subscribes or unsubscribes.
* **Staging / Beta**: The buttons are not shown.
* **Development / Local copy**: The buttons are shown on the "My account" page, but do not send any requests to Mailchimp.

---

Screenshot 2020-03-24 at 17 50 35
Screenshot 2020-03-24 at 17 51 02


marliekeverweij:
I've added Dutch translations. I've also removed 'free' and 'e-mail' from the English translations as I think it obvious that the newsletter is free and send by e-mail. But if it's better to add it (so exactly users know what they sign up for), I am fine too :)

PR #3295, opened 2020-03-30
Tweak title of "peak load" and "mean load" in chart date picker

antw:
We recently added text to the "Select date" field for hourly charts to include "peak daily load" or "mean daily load". This unfortunately makes no sense in at least one chart: the electricity price per hour.

This PR adjusts the wording slightly:

* peak daily load → daily peak
* mean daily load → daily mean

@marliekeverweij Would you mind double-checking the NL translations please?

### DIFF
--- a/config/locales/en_output_elements.yml
+++ b/config/locales/en_output_elements.yml
@@ -37,8 +37,8 @@ en:
       read_more: "read more"
       empty: "There is no data to display for this chart"
       whole_year: "Whole year"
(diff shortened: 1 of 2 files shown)


marliekeverweij:
Dutch translations are correct :) 

BTW the title of this PR suggests that the content is different

antw:
> BTW the title of this PR suggests that the content is different

Oops. GitHub inferred that from a commit which was wrongly included in this PR, which I then removed.

Thanks for checking the translations!

PR #3298, opened 2020-04-03
Compile input yamls

sfp-kemp:
We can use multiple yaml files to constuct a single model. This was requested by @michieldenhaan to make sure we have less merge conflicts in the future. 
This merge only aim to make it possible to compile yaml files and doesn't actually split the file. The plan is to pick that up once @noracato is done with migrating the descriptions.


sfp-kemp:
Don't mind the strange history. I branched off at the wrong place but that should be fixed once we squash it.

lib/ymodel/lib/ymodel/loadable.rb
```diff
@@ -21,11 +21,40 @@ def schema
...
+      @index = key
+    end
+
+    def index
+      @index || :id
+    end
+
     protected
 
+    # I dislike the format of this method a lot. Rubocop made me do it..
```
sfp-kemp, line 43: I really dislike the way this method is factored. I would love to get suggestions about how to improve it without getting scolded by the linter .

lib/ymodel/lib/ymodel/loadable.rb
```diff
@@ -21,11 +21,40 @@ def schema
...
-        .map(&:symbolize_keys)
+      @records ||=
+        if compiled?
+          Dir.foreach(source).reduce([]) do |memo, filename|
+            if filename.match?(/\A.*(.yaml|.yml)\z/)
+              memo + YAML.load_file(File.join(source, filename))
+            else
+              memo
+            end
+          end
```
antw, line 53: Personally, I don't think we should allow both ".yaml" and ".yml" extensions; it's reasonable to expect files to have a consistent naming convention. Elsewhere in ETModel (and other projects) it's always ".yml". The `_source` method for non-compiled YModel files still expects only ".yml".

```suggestion
          Dir.glob(File.join(source, '*.yml'))
            .flat_map { |name| YAML.load_file(name) }
```

sfp-kemp:
@antw can I merge this?

antw:
> can I merge this?

Yeah, looks good to me! :+1:

PR #3299, opened 2020-04-03
Migrate output elements

noracato:
This is a big one, because the output elements intertwined with a lot of other stuff. I already made them index on keys, however, the id's of the output_elements have to stay until we migrated the constraints to ymodel.

I also had to make some minor adjustments to ymodel itself. Like changing the default foreign keys and some error raising.

The keys for the output_element_series are new. I created them by adding the label and output_element_key together.


app/controllers/output_elements_controller.rb
```diff
@@ -19,7 +19,7 @@ def batch
     ids = params[:ids].to_s.split(',').reject(&:blank?).uniq
 
     json = OutputElementPresenter.collection(
-      OutputElement.where(id: ids),
+      OutputElement.where(key: ids),
```
sfp-kemp, line 22: I think this is a bit confusing. The variable is called ids in the params and as variable but used to hold keys.

noracato: Agreed. This happens in more (actually a lot of) places, also in the chart js code, that's why I was a bit reluctant to change all ids into keys. It might take me a few hours to sort it all out. But I totally agree that it can be confusing to read `id` everywhere when we are talking about keys. I can start on changing it right now in this branch, or merge this and create a new ticket?

sfp-kemp: > This happens in more (actually a lot of) places, also in the chart js code, that's why I was a bit reluctant to change all ids into keys. It might take me a few hours to sort it all out. 

I don't think you should fix every occurence in the codebase at this time but this PR will add couple of lines diff of the future fix. Fixing this controller and its routes shouldn't take an hour and will make sure you wont have to check over this file in the future.

noracato: This is now changed

app/models/output_element.rb
```diff
@@ -4,24 +4,22 @@
...
+#  show_point_label           :boolean
+#  growth_chart               :boolean
+#  max_axis_value             :float
+#  min_axis_value             :float
+#  hidden                     :boolean          default(FALSE)
 #
 
 # Entity used for filling charts
-class OutputElement < ActiveRecord::Base
+class OutputElement < YModel::Base
```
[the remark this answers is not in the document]
noracato: ...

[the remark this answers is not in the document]
noracato: Okay, so comments don't count...

lib/tasks/ymodel.rake
```diff
@@ -61,11 +61,10 @@ namespace :ymodel do
         referenced_records.find do |ref|
           ref['id'] == record[old_foreign_key]
         end
-
-      next unless related.present?
-
+      # Some relations are optional
+      new_foreign_val = related.present? ? related[new_index] : nil
```
sfp-kemp, line 65: It seems strange to me that we would want to cast `''` to nil. Is there a good reason to do this in the general case? It muddies up the code and might introduce unexpected behaviour.

spec/controllers/descriptions_controller_spec.rb
```diff
@@ -3,44 +3,25 @@
...
-      expect(response.body).to_not be_empty
-    end
-
-    it 'renders nothing if no description exists' do
-      get :show, params: { id: -1 }
-
-      expect(response).to be_successful
-      expect(response.body).to be_empty
-    end
-  end
```
sfp-kemp, line 25: Why did these tests dissapear? Is this behaviour no longer required?

config/routes.rb
```diff
@@ -98,11 +97,11 @@
...
-  resources :output_elements, only: %i[index show] do
+  resources :output_elements, param: :key, only: %i[index show] do
     collection do
-      get 'visible/:id',   action: :visible
-      get 'invisible/:id', action: :invisible
-      get 'batch/:ids',    action: :batch
+      get 'visible/:key',   action: :visible
+      get 'invisible/:key', action: :invisible
+      get 'batch/:keys',    action: :batch
     end
```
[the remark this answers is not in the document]
sfp-kemp: Apperantly I'm not good at using github..

app/views/descriptions/show.html.haml
```diff
@@ -1,18 +1,14 @@
 #overlay_container
-  - if @description.title
-    %h1= raw t(@description.title)
+  %h1= I18n.t(@chart.title_for_description)
 
   %p
-    - if @description.content
-      = raw @description.content
+    - if @description.blank?
+      No description available yet
```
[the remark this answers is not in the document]
noracato: The descriptions for charts are optional. You are right about the fact that this shouldn't exist. In fact I can just remove the whole if statement, because the check for redendering the info button on a chart has become more strict.

lib/ymodel/lib/ymodel/loadable.rb
```diff
@@ -10,6 +10,7 @@ module Loadable
     # overwrite the default ymodel filename associated with that model.
     def source_file(filename)
       @source = filename
+      @records = nil
```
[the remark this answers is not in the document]
noracato: Because I wanted the fixtures used as source files for YModel classes like OutputElements. These classes were already initialised/triggered, so I retriggered them with the fixture files.
Maybe you know a better way to point the YModel classes to the fixtures folder?

spec/controllers/descriptions_controller_spec.rb
noracato: The show_description route is no longer used in the code base. I removed it along with its specs.

lib/ymodel/lib/ymodel/loadable.rb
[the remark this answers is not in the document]
sfp-kemp: I'm not to sure why you would want to add fixture files. Could a mock isolate the test you are trying to write?

lib/tasks/ymodel.rake
noracato: It looked strange to me to see this in a slide:
```
output_element_key: ''
alt_output_element_key: ''
```
instead of 
```
output_element_key:
alt_output_element_key:
```
When there is no related element.
But I agree it muddies up the code, and all relations will work regardless.

sfp-kemp: I thought you agreed that this should be deleted. Could you still remove the change?

PR #3304, opened 2020-04-09
removed old stuff from the sign up page

sfp-kemp:


app/views/layouts/form_only.html.haml
```diff
@@ -20,10 +20,11 @@
 
   %body#form-only{ class: I18n.locale }
     %header#top
-      %a{ href: root_path }
-        %img(src="/assets/header/logo400-blue@2x.png" width="400" height="49" alt="Energy Transition Model")
+      .container
```
antw, line 23: Personally, I think I preferred the logo center-aligned. It doesn't look quite on the left. The gears in the 400px logo are slightly too wide, which causes "Energy Transition Model" to be misaligned with the content inside the "Create your account" container.

app/views/users/new.html.haml
```diff
@@ -1 +1,21 @@
...
+
+      .row
+        = f.input :password, required: true
+        = f.input :password_confirmation, required: true
+
+    .section.highlight
+      .row
+        = f.input :company_school
+
+    = f.input :allow_news
```
antw, line 19: `allow_news` checkbox should only be shown in environments where it is supported (i.e. not staging).

```suggestion
    - if MailchimpService.configured?
      = f.input :allow_news
```

config/locales/en_users.yml
```diff
@@ -44,19 +44,6 @@ en:
...
-      unsubscribe: Unsubscribe
-      subscribed: |
-        You are subscribed to our newsletter. We send roughly four
-        e-mails a year with information about improvements and changes to the
-        model.
-      not_subscribed: |
-        We occasionally publish a newsletter with information about
-        improvements and changes to the ETM. We send roughly four messages a
-        year and you can unsubscribe at any time.
-
```
antw, line 59: These texts are used on the "My account" page.

sfp-kemp:
Won't merge this for now.

PR #3316, opened 2020-04-16
Fix blockcharts by using keys for output_element_series

noracato:
@michieldenhaan pointed out the blockcharts were broken. I removed all references to ids and exchanged them for keys.


sfp-kemp:
@noracato I'm not sure what this does. It seems to conflict with with master and I don't understand how this mr relates to @antw's last commit. Maybe its best if he reviews this?

noracato:
Ah I see @antw already fixed this in his last commit, just after I put this PR up :) I'll close this one!

PR #3359, opened 2020-06-02
Change "max" chart downsample to use all series

### LINKED ISSUE #2867: Should 'whole year' view of hourly charts show daily peaks or daily means?
The yearly overview of the net load charts (lv, mv and hv) is counter intuitive. It shows daily average values while users mainly look at the peaks. It would be a great improvement if we could show the daily max values in these charts.  

Will this be an easy fix @antw ? 

screen shot 2018-09-11 at 15 51 21

antw:
Changes the "max" downsampling method for hourly charts to instead select the maximum cumulative load in each 24-hour period, and show the loads of each serie in that hour, rather than selecting the maximum of each series in any hour.

It also sets the dynamic demand chart to use "max" downsampling. This affects the annual views of:

* Demand curve of hybrid heat pump in households
* Electricity demand per hour
* Electricity demand on high voltage net
* Electricity demand on low voltage net
* Electricity demand on medium voltage net
* Hourly heat demand for district heating
* Hourly heat production for district heating
* Hydrogen demand / production / storage
* Network gas demand / production / storage
* (Seasonal) storage of heat
* Solar PV curtailment

**This is quite a lot of changes!** I'm not sure it makes sense for all these charts to show the "max"/peak load in the annual view. I'd be fine delaying this until we support per-chart interpolation settings.

@marliekeverweij What do you think?

Closes #2867

### DIFF
--- a/app/assets/javascripts/d3/dynamic_demand_curve.coffee
+++ b/app/assets/javascripts/d3/dynamic_demand_curve.coffee
@@ -45,6 +45,8 @@ seriesOpacity = (visibleSeries) ->
 
 D3.dynamic_demand_curve =
   View: class extends D3YearlyChartView
+    downsampleWith: 'max'
+
     initialize: ->
       D3YearlyChartView.prototype.initialize.call(this)
 
--- a/app/assets/javascripts/d3/merit_transformator.js
+++ b/app/assets/javascripts/d3/merit_transformator.js
@@ -1,6 +1,7 @@
 /* globals d3 */
 
-var MeritTransformator = (function () {
+// eslint-disable-next-line no-unused-vars
+var MeritTransformator = (function() {
   'use strict';
 
   function downsampleFunc(name) {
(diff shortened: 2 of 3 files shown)


antw:
@marliekeverweij Even if you like this, please hold off on merging until I've resolve the Hound errors. They're warning about a possible bug which I want to look into first.

antw:
I've resolve the issues with this PR. If you prefer the changes over the status quo, you're welcome to merge.

marliekeverweij:
Great that you have changed the "max" downsampling method for hourly charts!

My proposal would be to only show the daily max for the electricity network charts, as these charts should really show the daily load peaks!
82797919-12638e80-9e78-11ea-89c1-25125b245601

For other hourly demand charts (for example 'Electricity demand per hour') and other hourly charts there was still a discussion going on in issue https://github.com/quintel/etmodel/issues/2867 if we should show the daily mean of daily max in the 'whole year' view. I think that nothing has been concluded yet, unfortunately!

antw:
I _think_ the conclusion is that it's not right to apply the "max" downsampling to all the charts I mentioned in the original post. We should close this, unmerged.

Ideally we'd add a way to specify the downsampling method for each chart in the YAML, rather than forcing it to be the same for all charts of the same time.

For example:

```yaml
- key: chart_one
  type: demand_curve
  display_options:
    downsample_with: mean
  # ...

- key: chart_two
  type: demand_curve
  display_options:
    downsample_with: max
  # ...
```

PR #3363, opened 2020-06-05
Multiple electricity interconnectors

antw:
* Adds slides and sliders for interconnectors 2 through 6.
* Supports minor changes in the ETEngine API.
* Makes adjustments to charts and tables to show the new interconnectors.
* Adjusts sankey charts to hide series with no demand.
* Adds the excellent "Select curve from a scenario" feature written by @noracato.


marliekeverweij:
Looks good! Great work and the scenario selection works flawlessly! Haven't tested super thoroughly, but @michieldenhaan and @AlexanderWirtz will do that next week on beta

PR #3407, opened 2020-06-16
Separates demand series on "Electricity production"

antw:
Adds a "Demand (domestic)" series which does not include export, shown only for the future. Please merge https://github.com/quintel/etsource/pull/2308 first.

I'd appreciate a quick sanity check on [the "Demand (domestic)" translation](https://github.com/quintel/etmodel/blob/0ce5730dd639ec192be8e05db2972ec6605cece1/config/locales/interface/output_element_series/nl_labels.yml#L215). "binnenlandse" is used elsewhere in the ETM, but Google thinks it should be "huiselijk" (I'm guessing this refers to a house or household, rather than an area?).

Screenshot 2020-06-16 at 10 30 42

### DIFF
--- a/config/interface/output_element_series/source_of_electricity_production.yml
+++ b/config/interface/output_element_series/source_of_electricity_production.yml
@@ -132,8 +132,20 @@
   output_element_key: source_of_electricity_production
   key: solar_source_of_electricity_production
 - label: demand_incl_export
-  color: "#FF0000"
+  color: "#5D59C7"
   order_by: 14
+  group: ''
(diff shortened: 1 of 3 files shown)


ChaelKruip:
> I'd appreciate a quick sanity check on the "Demand (domestic)" translation. "binnenlandse" is used elsewhere in the ETM, but Google thinks it should be "huiselijk" (I'm guessing this refers to a house or household, rather than an area?).

Correct. "binnenlanse" is the correct term. 👍

PR #3422, opened 2020-06-29
Saved scenario edit button

noracato:
Hi!
I think I'm done :) Not too sure about the title size/placement in 'show' mode though. I wanted to upload some screenshots here, but github seems a little confused this morning.


antw:
On second thoughts, I think the padding you chose for the buttons (10px 15px) is better than my suggestion. It prevents the button from overwhelming the rest of the content:

Screenshot 2020-06-29 at 15 26 55

BTW, I'm *not* saying you must do this. I'm just trying to spark some ideas for improving the design of the page.

noracato:
Thanks so much for your suggestions! They definitely sparked something 😄 

I especially love the flag and making the edit-button into a smaller link. I'll go ahead and play with it a bit!

noracato:
This is what I came up with:

Screenshot 2020-06-30 at 14 34 08

And this one for the edit form:

Screenshot 2020-06-30 at 14 56 49

As you can tell I liked your comment about tying in with the landing page some more. I thought continuing the style from the landing page on this page would make the non-play part of the ETM seem more like a whole. Curious to hear what you think!

Also, I changed the warning at the bottom. I agree the message that was there didn't contribute to anything. Maybe this is better? I fine with leaving it out if you think that's best.

antw:
That looks amazing; huge improvement!

I've only a few very minor comments:

* I'm unsure about greying out the "Netherlands 2050" part; is that not important context about the scenario?
* I still think the buttons would look a little better with a border. On the front page we use `box-shadow: inset 0 0 0 1px rgba(0, 0, 0, 0.1)`.
* Is the green button a different colour than on the front page? It looks a bit bright. The "Discard changes" button text is also a little too light IMO; it might confuse some users into thinking the button is disabled.

Nitpicking aside, I love the changes you made. :tada:

noracato:
The buttons now have borders, I agree that it makes them more visible. And I made the area/year part darker as well.

About the green; I copied the exact css from the landing page, so it should be the same color. We can always dull it down a bit on this page if you'd like that!

Screenshot 2020-07-02 at 10 43 01

antw:
> About the green; I copied the exact css from the landing page, so it should be the same color.

Oh fair enough! Maybe the yellow background on the root page makes it look less saturated? Either way, ignore my previous comment then; I think we should leave it as it is.

noracato:
😄 

Is it okay if I squash-merge this? Or do want to take a last look at the code?

PR #3426, opened 2020-07-02
Add blank flag for Northern Ireland

marliekeverweij:
Screenshot 2020-07-02 at 10 47 13


antw:
I don't really like this: a blank flag will misalign a number of elements in the website where a flag is expected. I'll see if I can come up with an alternative which doesn't need a blank image.

marliekeverweij:
Ah good one, hadn't thought of that! It will be temporary though (until August) and only on our beta server. It is needed for the presentations and demos that we will give to other institutions within NI in July & August.

antw:
https://github.com/quintel/etmodel/commit/3ec051967999b256d805e2f19e3666f3eb419caf

Issue #3428, opened 2020-07-03
Scenario dropdown in interconnector-price-scenario-selector could be structured more


marliekeverweij:
As mentioned by @AlexanderWirtz and @ChaelKruip during the wrap-up meeting for interconnectors this week, I would be really nice if the scenario dropdown in interconnector-price-scenario-selector would be structured more. At the moment it shows all scenarios that are made by the user and this could get quite full. Some ideas to improve this:

- [x] One of the first things we could do is to only show the scenarios with the same end year as the scenario currently working on -> DONE!

- [ ] Another improvement would be to only show scenarios of neighboring regions. This would require a lot more effort as we than have to determine the neighboring regions for each dataset!!

- [ ] When the ['archive scenario'-function](https://3.basecamp.com/3797409/buckets/13985863/todos/2236539777) is implemented, then the drop down will also get more structured :) 

Do people have more ideas?

marliekeverweij:
I think we could close this issue as @noracato has made a big improvement :) 

- Not showing scenarios that have the same area code as the current scenario
- Putting the scenarios that have the same end year as the current scenario at the top of the list
- (Sub)order on last updated

PR #3430, opened 2020-07-03
Interconnectors: Only choose from scenarios with same end year

noracato:
See #3428

### DIFF
--- a/app/controllers/saved_scenarios_controller.rb
+++ b/app/controllers/saved_scenarios_controller.rb
@@ -14,7 +14,10 @@ class SavedScenariosController < ApplicationController
         if current_user
           render json:
             SavedScenarioPresenter.new(
-              current_user.saved_scenarios.order('created_at DESC')
+              current_user
+                .saved_scenarios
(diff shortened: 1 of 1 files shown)


marliekeverweij:
For 2050 it works! However, for 2030 I also see 2050 scenarios in the list.... 

Screenshot 2020-07-03 at 15 24 21

michieldenhaan:
Does this mean that users can now _only_ select scenarios with the same end year? If so, I'm not sure if that's the way to go. It can be quite confusing why only certain scenarios show up and in some cases it can make perfect sense to select a price curve from scenarios with different end years. I would be in favour of changing the scenario _order_ (e.g. first show scenarios with the same end year) but not hiding scenarios all together.

ChaelKruip:
> Does this mean that users can now only select scenarios with the same end year? If so, I'm not sure if that's the way to go. 

I agree with @michieldenhaan! We should not hide/remove anything just yet.

> I would be in favour of changing the scenario order

Yes! 👍

PR #3433, opened 2020-07-03
Switch to system font stack

antw:
We already use the system font stack on a number of ETM pages (notably the root page where you select whether to continue a scenario, start a new one, or start a preset). This pull request changes the scenario pages to also use these fonts.

Before pushing ahead I'd like to ensure @quintel/core is on board.

> Defaulting to the system font of a particular operating system can boost performance because the browser doesn’t have to download any font files, it’s using one it already had. That’s true of any “web safe” font, though. The beauty of “system” fonts is that it matches what the current OS uses, so it can be a comfortable look.
>
> *― [System Font Stack](https://css-tricks.com/snippets/css/system-font-stack/), CSS Tricks*

In our case, the pages won't load any faster since we weren't using a web font, but the system stack does look more moder and, IMO, is a little easier to design for.

### Comparisons

| Before | After |
| --- | --- |
| 0-before | 0-after |
| 1-before | 1-after |
| 2-before | 2-after |

### Full-size examples

Screenshot 2020-07-03 at 17 18 35
Screenshot 2020-07-03 at 17 18 53

### DIFF
--- a/app/assets/stylesheets/chart.sass
+++ b/app/assets/stylesheets/chart.sass
@@ -290,7 +290,7 @@ div.d3_container
     svg
       shape-rendering: crispEdges
   text
-    font-family: $body_font_family
+    font-family: $body-font-family
     font-size: 12px
   .year.small
     fill: #777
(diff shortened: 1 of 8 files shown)


marliekeverweij:
Very nice! Like it a lot!

PR #3435, opened 2020-07-07
Custom curves scenario order: same end year first, exclude same area

noracato:
In #3430 there was some discussion about _not_ showing scenarios with the same end year. I changed the order into:

- Not showing scenarios that have the same area code as the current scenario

- Putting the scenarios that have the same end year as the current scenario at the top of the list

- (Sub)order on last updated


marliekeverweij:
Great! Yes, it's better to show the scenarios with the same end year on top and the others below. 

I am not sure about not showing scenarios with the same area code. It sounds logical though!

PR #3444, opened 2020-07-23
Add featured scenarios, replacing presets

antw:
Adds the ability for admins to "feature" a scenario. Featured scenarios appear on the front page under "Start an existing scenario" where presets were previously located.

This PR:

* Adds a migration that creates the featured scenarios database table and model.
* Creates a new info@energytransitionmodel.com admin user, with a random password.
* Fetches the list of presets from ETEngine, creates a new ETEngine scenario for each one, then creates a saved scenario and featured scenario.
* Adds links and forms for creating a featured scenario.

A featured scenario is created by clicking "Feature this scenario" from the saved scenario page (available only to admins):

Screenshot 2020-07-23 at 15 14 20

This shows a form allowing the user to set:

* Group - the group in which the featured scenario will appear on the front page.
* Custom English and Dutch titles and descriptions.

I haven't loaded the featured scenario form in-place, like the "Edit title and description" feature does, as this will only be used by admins. I didn't feel the extra effort was worth it.

When a scenario has been featured, the custom English or Dutch titles and descriptions will be shown rather than the title or description which was set on the saved scenario.

A featured scenario may be edited:

Screenshot 2020-07-23 at 15 16 11

This allows setting new titles or descriptions and changing the group.

At the bottom of the page, the scenario may be "unfeatured". The user is warned that the custom localized titles will be removed, and the scenario will revert back to the old settings. This is to allow them the opportunity to copy one of the localized descriptions prior to deleting.

Screenshot 2020-07-23 at 15 47 10

---

Some translations are needed before merging. Please and thank you!

* ["Feature this scenario" links](https://github.com/quintel/etmodel/blob/bb21ee14ffd624017783b707657350c062b1a0de/config/locales/nl_etm.yml#L212-L213)
* ["Unfeature this scenario" information and warnings](https://github.com/quintel/etmodel/blob/bb21ee14ffd624017783b707657350c062b1a0de/config/locales/nl_etm.yml#L222-L232)
* [Featured scenario form labels](https://github.com/quintel/etmodel/blob/bb21ee14ffd624017783b707657350c062b1a0de/config/locales/nl_simple_form.yml#L8-L13)
* [Featured scenario form buttons](https://github.com/quintel/etmodel/blob/bb21ee14ffd624017783b707657350c062b1a0de/config/locales/nl_simple_form.yml#L40-L44)

---

I haven't removed presets from ETEngine, Atlas, or ETSource yet. We should do this eventually, but migrating the presets into saved scenarios require that they still be available in ETEngine.

Including @ChaelKruip @marliekeverweij @MartLubben.


noracato:
Very nice! I really like it! 🎉 

I added the Dutch translations and had a quick spin with the new feature. It works great. One little nitpicky thing, when I'm newly featuring a scenario the 'unfeature' button still appears in the top right of the form. Not sure if that's worth doing something about as the features only for admins 😄  Once again, nice work

🎉

antw:
> One little nitpicky thing, when I'm newly featuring a scenario the 'unfeature' button still appears in the top right of the form.

Good catch. I've fixed it. 👍

PR #3446, opened 2020-07-27
Style Scenarios#index like #show and root-page

noracato:
In view of the deploy tomorrow, I thought it would be nice to tie Scenarios#index in with its neighbours (Saved)Scenario#show, #feature and the root page.

Old look:

Screenshot 2020-07-27 at 15 06 54

New look:

Screenshot 2020-07-27 at 15 05 17

I did nothing with the avaible options (compare, merge, compare and combine), this PR is just a small restyle. @antw feel free to comment on the design or to add your own changes 🙂

### DIFF
--- a/app/assets/stylesheets/scenarios.sass
+++ b/app/assets/stylesheets/scenarios.sass
@@ -142,14 +142,13 @@ table.default
           font-size: 1rem
           display: inline-block
           width: 100%
-
-          &:hover
-            color: darken($landing-blue, 4%)
-            text-decoration: underline
         .info
           color: black
         .author
           color: $landing-grey
+        &:hover .title
+          color: darken($landing-blue, 4%)
+          text-decoration: underline
 
       #scenario_paginator td
         border: none
--- a/app/assets/stylesheets/scenarios.sass
+++ b/app/assets/stylesheets/scenarios.sass
@@ -31,10 +31,7 @@ table.default
   height: max-content
 
 .scenario_wrapper
-  &#index
-    padding: 5px 20px 110px 20px
-
-  &#show_scenario, &#feature_scenario
+  &#show_scenario, &#feature_scenario, &#index
     width: $landing-width
     margin: 0 auto 3rem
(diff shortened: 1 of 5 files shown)


marliekeverweij:
I like the new style!

The new style does attract more attention to the buttons IMO and there is some discussion about the buttons: whether we should keep all of them. I do not think we should have that discussion now, as the functionality behind the buttons still works (I just quickly checked it), right? I will write a ticket for our backlog to have that discussion at another moment in time. What do you think? @ChaelKruip @antw @jorisberkhout

noracato:
Thank you for your comments @antw

I was also wondering what to do with the 'save now' button. I came up with this, but the button might still be a little prominent. (Also added that when you don't have a current scenario, the button does not appear)

Screenshot 2020-07-27 at 17 32 56

antw:
> I was also wondering what to do with the 'save now' button

I got rid of it completely in my screenshots. It's not difficult to figure out where to place the button if we get rid of it. 😆 

If we were to add [the new save button and scenario subnav soon](https://3.basecamp.com/3797409/buckets/16975857/todos/2794803210#__recording_2811582138), I think it's okay for us to remove the "Save" link from the saved scenario list. If it's going to be some time before we do that, perhaps it would be preferable to keep it?

My concern with having it next to the "Compare / Merge" buttons, is that those other buttons are all related to the checkboxes, but "Save" is not.

We could do something like this:

save-placement

Moving it to the right means that the buttons related to the checkboxes still appear close to the checkboxes, and the extra border before the "Save" buttons helps show that it is distinct from the others. It still feels a bit out-of-place to me though.

noracato:
> I got rid of it completely in my screenshots. It's not difficult to figure out where to place the button if we get rid of it. 😆

Haha, no then it's not 😄 

I find that wherever I place the button, it doesn't feel right. But I also think we should keep it there until we create the new subnav. I think your solution/placement is the lesser evil (putting it on the right, with the separator).

PR #3462, opened 2020-09-15
Update text of heat storage costs and losses slider

marliekeverweij:
Screenshot 2020-09-15 at 13 30 33
Screenshot 2020-09-15 at 13 36 21


michieldenhaan (commented):
We don't take into account the heat loss compensation costs right? As these are already included in the ETM calculation when balancing supply and demand. I think we should say this explicitly in the text, otherwise it's hard to translate the Ecovat cost chart to ETM slider value

michieldenhaan (approved):

PR #3465, opened 2020-10-05
Make it possible to create a scenario based on an ESDL file

noracato:
Hi @antw, thanks for your work on this branch! Looks great :)
Creating this PR kept falling to the bottom of my to-do list - but here we are.


antw (approved):
Looks and works great. Thanks @noracato!

PR #3470, opened 2020-10-16
Add secondary scenario navigation

antw:
This PR adds a new scenario navigation bar on the main scenario page. This bar contains the current dataset name, end year, and saved scenario name (if applicable).

~[Some translations are needed](https://github.com/quintel/etmodel/blob/976dd7342bb55b1a08dcfef0adce94a3e26f7ab2/config/locales/nl_users.yml#L54-L62) please.~

Screenshot 2020-10-16 at 14 21 40
Screenshot 2020-10-16 at 14 22 42

On the right side of the bar is a "Save Scenario" button, and "Actions" which hides less-frequently-used features. The dropdown menus are all subtly animated.

If also features a refreshed main header:

* Reduced in size.
* Colours adjusted to match the redesigned ETCentral.
* Visitors can more easily access their saved scenarios.
* Language selection moved to the main header.

Clicking "Save Scenario" does one of two things:

* If the user is already existing a saved scenario, it immediately saves the updates and shows a little notification.
  Screenshot 2020-10-16 at 14 28 10

* If the scenario has not been previously saved, the visitor will be prompted to enter a title and optional description, and their saved scenario will then be created.

Finally, I removed the login form from the header and redesigned the main login page:

Screenshot 2020-10-16 at 14 23 46


michieldenhaan:
This is awesome @antw!

ChaelKruip (approved):
Great work!

PR #3471, opened 2020-10-20
Migrate area codes of saved Gelderland scenarios

antw:
Updates the area_code of saved scenarios associated with Gelderland scenarios. See quintel/etengine@5739e63f.


michieldenhaan (approved):
Thanks! When I run this locally it says 0 rows updated for Lingewaard. But that's presumably because scenarios are missing in my local copy of the production DB?

antw:
> that's presumably because scenarios are missing in my local copy of the production DB?

Yes indeed. The real result will be:

```
[...]
-- om_lingewaard -> GM1705_lingewaard
   -> 0.0067s
   -> 8 rows
[...]
```

PR #3501, opened 2020-11-12
Remove InputElementView smartStep

antw:
Input value step values would be overridden whenever there were insufficient pixels to display each value discretely. If the slider is 100px wide and represents values between 0 and 1000 and the step value was 0.1, the step value would be overridden with a value equivalent to 1px (10).

This behaviour has been removed, and the values set in the interface YAML files will be trusted.


michieldenhaan (approved):
Thanks, looks perfect!

image

Issue #3500, opened 2020-11-12
Auto scaling of slider units


michieldenhaan:
The ETM includes quite a few sliders with 'absolute' units. E.g. power plant capacity in MW, captured CO<sub>2</sub> in Mton, imported heat in PJ etc. For some smaller areas in the ETM, these units are too big to be meaningful. It would be nice if the slider units automatically scale to smaller units for those areas (capacities to kW, CO<sub>2</sub> to kT, PJ to TJ or GJ etc.)

Geoneer:
#3505

michieldenhaan:
I noticed that in datasets with less than 50k inhabitants, the population slider does not work. I think this is related to the start value, which is 0.0 million (the slider unit is millions and the step value is 100k), causing the slider to break. If I remember correctly, this did work fine before.

@antw Do you see a quick fix for this?

![image](https://user-images.githubusercontent.com/32056448/100426178-bf6dcc80-3090-11eb-80ea-9c18de531d65.png)

antw:
> If I remember correctly, this did work fine before.

I thought at first this might be because we changed it so the configured step value is always respected, but reverting that commit didn't have any effect. I'm struggling to see where the step values may have been previously scaled; we do scale step values for small datasets in ETEngine, but ETModel doesn't use those step values (it configures its own, per-slider). It doesn't seem that the removal of the scenario scaling feature in ETModel is the culprit either.

I've sent a PR which will scale step values of sliders when the step is too large to be useful.

michieldenhaan:
Thanks! Both @redekok and I (vaguely) remember that we did have some form of auto-scaling of slider units not so long ago. E.g. we both remember local datasets with installed capacities in kW rather than MW. I also encountered a local scenario someone made via the front-end which has 130 TJ heat import. With the current slider (unit PJ, step size 0.1) this is impossible: when entering 0.13 PJ in the slider box it gets rounded to 0.1 PJ, so 100 TJ.

antw:
> E.g. we both remember local datasets with installed capacities in kW rather than MW.

You're right, and I found the code that does it. The code to scale the units from – for example – from MW to kW doesn't run anymore since we don't do anything with the scaling information sent by ETEngine (as it'll likely be removed once support for scaled scenarios is removed from ETE). As far as I can tell this information was only sent for scaled scenarios, not for smaller datasets.

We can enable this code for _all_ scenarios and datasets but this is more likely to have side-effects.

[...]

PR #3511, opened 2020-11-27
Scale slider steps when min and max range is small

antw:
In small datasets, the range between the minimum and maximum values of a slider may be too small for the step to be meaningful. For example, the population slider with a step of 100k is useless in a dataset with only 50k people.

This change automatically scales the step value of a slider, dividing by ten until there are at least 20 "steps" between the minimum and maximum slider value. This should keep the step value in small datasets meaningful, without making it _too_ small.

Ref #3500

### DIFF
--- a/app/assets/javascripts/lib/models/input_element.coffee
+++ b/app/assets/javascripts/lib/models/input_element.coffee
@@ -90,6 +90,22 @@ class @InputElement extends Backbone.Model
     if @get('key') == 'settings_enable_merit_order'
       App.update_merit_order_checkbox()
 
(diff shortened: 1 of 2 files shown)


michieldenhaan (approved):
🎉

michieldenhaan:
Do you think we can cherry-pick this to PRO (as the sliders are broken there as well)? Or do you prefer to wait a bit to see if this has any unforeseen fall-out?

antw:
> Do you think we can cherry-pick this to PRO

Done. We can always revert if there are any unintended consequences.

PR #3512, opened 2020-11-27
Scale sliders to an appropriate unit

antw:
Scale all sliders based on their mean value (the value of the slider when at 50%). This restores scaling of sliders for small datasets, while ensuring reasonably consistent results for sliders in a group (assuming they have similar mean values).

This change will affect the display of sliders for all datasets, I don't expect major fallout, but I think it would be wise for it to be tested and spend time on beta before moving it to production.

Ref #3500


michieldenhaan (approved):
🥇

PR #3516, opened 2020-12-11
Rename / remove old Noardeast Fryslan municipalities

antw:
Counterpart to https://github.com/quintel/etengine/pull/1145. Renames the area code of saved scenarios and transition path charts for three Noardeast Fryslan datasets.


michieldenhaan (approved):
Thanks! Forgot about this one 🙈

PR #3517, opened 2020-12-14
Hide remaining households hot water geothermal node

mabijkerk:
The table _Final demand heating and hot water_ in residences showed a duplicate entry for _Ambient heat (for hot water)_ as shown in the screenshot below. This entry was not affected by any of the sliders in the relevant section of the model. Further inspection showed that it represented the gquery _geothermal_used_for_hot_water_in_households_. The other geothermal gquery in output_element_series was set to always hidden so I've updated this gquery to match it.

Screenshot 2020-12-14 at 09 37 57


michieldenhaan (approved):

ChaelKruip:
Nice catch @mabijkerk!

PR #3525, opened 2021-01-04
Additional outdated output element series found removed for etmodel #3517

mabijkerk:
The gquery geothermal_used_for_hot_water_in_households is outdated since geothermal heat has been coupled to the heat network, instead of directly to the households. This issue was already addressed in etmodel pull request #3517, but this output_element_series had not yet been deleted.

Since there is no slider that has an effect on the geothermal heat used for hot water in households, this output_element_series has been removed.

### DIFF
--- a/config/interface/output_element_series/source_of_hot_water_in_households.yml
+++ b/config/interface/output_element_series/source_of_hot_water_in_households.yml
@@ -95,15 +95,3 @@
   dependent_on: 
   output_element_key: source_of_hot_water_in_households
   key: ambient_source_of_hot_water_in_households
-- label: geothermal
-  color: "#FF8400"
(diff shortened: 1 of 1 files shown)


mabijkerk:
@michieldenhaan In addition to the outdated geothermal heat issue addressed above, I noticed that the entire source_of_hot_water_in_households.yml file is outdated and set to always_hidden in the demand_households.yml. This is because both hot water and space heating have been combined in the source_of_heat_used_in_households.yml file.

Do you have a preference for deleting the source_of_hot_water_in_households, or leaving it as is?

michieldenhaan (approved):
@mabijkerk let's kill it!

PR #3537, opened 2021-01-11
Overhaul chart picker

### LINKED ISSUE #2837: Should the chart list have a search function?
As discussed in the [previous issue](https://github.com/quintel/etmodel/issues/2612) we thought that it would be nice to add a search-function to the chart list. Maybe in the upperright corner?

screen shot 2018-06-29 at 14 04 46

Let me know your opinions! @jorisberkhout @AlexanderWirtz @ChaelKruip @DorinevanderVlies @antw @MartLubben @michieldenhaan

antw:
* The list is now generated by a couple of Backbone views with JSON data returned from the ETModel API.
  * The list of output elements is fetched as soon as the user hovers the chart picker buttons, which means it appears faster than before. The list is also cached, and so appears instantly when re-opened.

* A search field has been added allowing users to quickly find the chart they want.
  * Fuzzy matching is supported: searching for "el prod" shows the "Electricity production", "CO2 emissions of electricity production", and "Electricity production per hour" charts.
  * Matching is also performed against group and sub-group names: searching for "flex" shows all charts in the "Flexibility" group.
 
* Charts which are already loaded will be greyed out and not selectable, to prevent the same chart being selected multiple times.

@marliekeverweij You requested this more than two years ago, so it's been a long time coming! It would benefit from [a double-check of the translations](https://github.com/quintel/etmodel/blob/9536cfb87669bc6fb2094081f1f8138d87793d2f/config/locales/nl_etm.yml#L30-L33) before merging.

101484495-a1864e80-3951-11eb-895f-ec9c46becebd

Closes #2837.

### DIFF
--- a/app/assets/javascripts/lib/models/chart_list.coffee
+++ b/app/assets/javascripts/lib/models/chart_list.coffee
@@ -298,26 +298,22 @@ class @ChartList extends Backbone.Collection
 
     # Launch the chart picker popup
     #
-    $(document).on "touchend click", "a.select_chart, a.add_chart", (e) ->
+    $(document).on "touchend click", "a.select_chart, a.add_chart", (e) =>
       e.preventDefault()
-      url = $(this).attr('href')
-      $.fancybox.open
-        autoSize: false
-        href: url
-        type: 'ajax'
-        width: 930
-        height: window.innerHeight - 100
-        padding: 0
-        afterShow: ->
-          # Pick a chart from the chart picker popup
-          #
-          $('#select_charts .select-chart').on 'touchend click', (e) =>
-            data_holder = $(e.target).closest('div.select-chart')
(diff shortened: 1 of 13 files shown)


marliekeverweij (approved):

noracato:
Looks amazing @antw! And works like a charm 🎉 

One thing that caught my eye is that the html tags in the group titles don't seem to work anymore. I think it's in the `.textContent` on `l:193` of ChartListView. We can fix that, or change the dutch title - as the example below seems to be the only one affected.
 
Other than that: great job

Screenshot 2021-01-11 at 17 18 09

antw:
> the html tags in the group titles don't seem to work anymore

Good catch. I changed this to use `innerHTML` as we know the text is safe to be treated as such.

noracato (approved):

PR #3548, opened 2021-01-19
Translations node properties table

marliekeverweij:
Goes with: quintel/etengine#1149


config/locales/en_node_details.yml
```diff
@@ -50,6 +58,10 @@ en:
     texts:
       'yes': 'Yes'
       'no': 'No'
+      d_w_s: day / week / season
+      i_h_d: immediate / hour / day 
+      i_h_d_w: immediate / hour / day / week
+      i_h_d_w_s: immediate / hour / day / week / season
```
antw, line 64: When I first saw these in a screenshot I found them pretty confusing. The reason being that the table already has entries like "€10 / full load hour" and "8 / year" where the "/" means "per": "€10 per full load hour" and "8 per year".

The result was that I had a hard time parsing what "immediate / hour / day" was supposed to mean. How would you feel about changing these to:

```suggestion
      d_w_s: daily, weekly, or seasonally
      i_h_d: immediately, hourly, or daily
      i_h_d_w: immediately, hourly, daily, or weekly
      i_h_d_w_s: immediately, hourly, daily, weekly, or seasonally
```

Ditto for the Dutch translations.

marliekeverweij: Good idea! I like it

marliekeverweij: I will update it

antw (approved):
LGTM! I think I've fixed all the reported issues, but it could do with a quick second check to ensure you're happy with my changes.

PR #3554, opened 2021-01-20
Add feedback option to main navigation

antw:
Adds a Stripe-style feedback form, allowing visitors to provide quick feedback either on the ETM in general, or on the current page.

Some metadata is included to help debug any issues the visitor is reporting and provide context to their message. A clone of the active scenario will be created.

Sorry to ask this again, but [there are translations needed](https://github.com/quintel/etmodel/blob/17c72c6299f68cc440f23bd098aeea1f25883996/config/locales/nl_etm.yml#L375-L383). 😇

Kapture 2021-01-20 at 15 07 17

Screenshot 2021-01-27 at 16 45 38


noracato (approved):
Very nice!! 🎉 

I added some nl translations (see image), @ChaelKruip feel free to change them.

Screenshot 2021-01-28 at 09 24 41

marliekeverweij:
I really like this feedback option!! The goal of this new feature is to improve the ETM, right?

What if users have questions about how something works? Should they then also use this button? I think that in that case it would be nice that they can also leave your contact details, so we can respond. We do have the [https://energytransitionmodel.com/contact](https://energytransitionmodel.com/contact) where you can leave your contact details, but that's a bit "far away" when you're working on a scenario. I would be in favour of adding some kind of "Question?" button aswell or changing it to a "Feedback/Question?" button. The goal of the "Question?" button is that we are more approachable to help out users when they're stuck, so they do not drop out.

PR #3563, opened 2021-01-25
End support for IE11

antw:
Visitors arriving a the ETM using IE11 or older will be greeted with a warning page advising them to upgrade their browser. They may still continue to the app, but should not expect everything to work. The warning message is only shown for IE users, since it's pretty much the only unsupported browser to appear in any significant numbers in our data.

Sorry to say that there are [some translations needed](https://github.com/quintel/etmodel/blob/a16fbed94f923fa7edfe4ed9c53fe69dacdeab85/config/locales/nl_etm.yml#L362-L372) before merging. I would like to have this merged before the feature freeze, but otherwise there's no rush.

Here's what it looks like in IE11 on Windows 10:

unsupported-ie11-win10

Safari is included on the page as it is one of the browsers we support, even though it isn't an alternative to IE (since there's no Windows version).


noracato (approved):
LGTM!

Another nice change to the ETM 🎉

PR #3564, opened 2021-01-25
New chart: Supply and demand peaks per month

marliekeverweij:


marliekeverweij:
@antw I moved the time frame information to the 'Other' section in the node properties table (instead of 'Technical') to show it less prominently. The consequence of that is that the 'Other section' moves to the top. Is it possible to keep the 'Other' section below the other two sections?

Screenshot 2021-01-25 at 16 00 50

mabijkerk (approved):
Looks fine to me!

antw:
> Is it possible to keep the 'Other' section below the other two sections?

Done in cda3d1e.

PR #3583, opened 2021-01-28
Add "What's new" section to the root page

### LINKED ISSUE #3562: Create a permanent link to the what's new page
Now that a deploy takes place every month, @noracato and I thought it would be a good idea if we would automatically put up the deploy-banner on PRO and also take it offline again.

I think 2 different banners should be implemented
1. Before the deploy, telling users the exact date and time when the deploy will take place so they know what to expect when. Start date banner: one week before the deploy. End date banner: deploy!
2. After the deploy, informing users that the deploy has taken place and directing users to the what's new page. I think this banner should only be shown if a significant update has taken place. So maybe this banner could be depended on if there are any new sections added to the What's new page.....
Start banner: deploy! End banner: one week after??

I actually think there should always be a link to the What's new page explaining the newest features, so we could also choose to add a permanent link.

What do other people think?

antw:
Screenshot 2021-01-28 at 16 38 20

@marliekeverweij Would you mind taking care of [the single translation needed](https://github.com/quintel/etmodel/blob/ef8e0d0e7c69fa869226db4b41cb7a5e8502843e/config/locales/nl_intro.yml#L27)? Please and thank you.

Closes #3562


marliekeverweij (approved):

marliekeverweij:
Great!!

PR #3584, opened 2021-01-28
Improve old scenario warning and add version number

### LINKED ISSUE #3581: Make warning box for scenario changes more subdued / less prominent
As dicussed in [this Basecamp discussion](https://3.basecamp.com/3797409/buckets/19561768/messages/3379099003) it would be nice if the warning box would be more subdued. This issue is about the lay-out and location of the warning box and less about the content (as the content is part of the flat tire).

Ideas:
- new location: at an equal position as the current 'Please note:...' 
- change the colour (not red)

Screenshot 2021-01-28 at 16 04 26

antw:
* ETModel::Version has been added which allows us to specify the most recent year, month, and day on which the ETM was updated.
* The "old scenario" warning now shows when the most recent change to the scenario was prior to the ETM release date.
* The warning will not show if the release date is in the future, which might be the case on beta and development.
* The warning will never show on featured scenarios. I think it looks unprofessional for featured scenarios to show a message about being out-of-date, and as *we* are the scenario creators we can be sure the slider settings and outcomes are as expected.

[Translations please!](https://github.com/quintel/etmodel/blob/07774f2591300cfe3e6807889c2dd458c6cd1e10/config/locales/nl_etm.yml)

K5zhzh

> Screenshot 2021-01-28 at 21 12 20

Closes #3581


michieldenhaan (approved):
🎉 epic!

ChaelKruip:
Great work! 🥇

PR #3590, opened 2021-02-01
Add warning to profile chart when lacking capacity

antw:
Adds a small warning message to the user profile chart whenever they select a technology whose curve is all zeros; this indicates that there was no demand, supply, or capacity installed. Based on feedback from Alexander on Basecamp.

There is one short translation required, which customises the message for this chart from the default ("There is no data for this chart") to something more meaningful. I worded the message differently from Alexander's suggestion – which was to mention the lack of capacity – since the technology *may* have capacity installed but still have no demand or supply.

Screenshot 2021-02-01 at 16 47 05


michieldenhaan (approved):
very nice, thanks! 💎 🚀

PR #3606, opened 2021-02-08
Add more context to the "outdated scenario" warning

antw:
I already dislike the wording of the outdated scenario warning message I added a week ago. It simply says "scenario settings may have changed" without explaining why.

Screenshot 2021-02-08 at 12 08 59

Given that this message is now shown much more frequently I think we need to explain ourselves better; and get across the point that many scenarios will be unaffected.

If you agree with the changes, there [are two short translations needed](https://github.com/quintel/etmodel/blob/95f3dc0edee0d78f6179262a860e3a41a3ac1358/config/locales/nl_etm.yml#L192-L198).

Here's what I propose:

**Default warning:**

Screenshot 2021-02-08 at 12 16 56

**Explanation expanded by visitor:**

Screenshot 2021-02-08 at 12 17 05

From the first paragraph I removed the message about sliders being changed: only a minority of scenarios are affected by most migrations. I also changed the colour from yellow to green, so that the message looks less alarming, as the notice is intended to contain useful information rather than suggest something is wrong.


ChaelKruip:
I like the changes you made!

One thing I feel a bit unsure about: when the user expands the "how does this affect the scenario" part, I would expect a message about how this particular scenario was affected. Which sliders have been moved? If any?

Is there a way to make that information available? If this is difficult (and I can imagine it is), perhaps we should rephrase the expanding text slightly to be more general: "How do updates of the ETM affect scenario's?" for instance.

antw:
> I would expect a message about how this particular scenario was affected. Which sliders have been moved? If any?

I very much agree, but at the moment we don't have this information available. We'd need an inventory of which inputs have been affected by migrations since the scenario was last saved. That would allow us to show a list of affected inputs.

To show the differences in input *values* would mean having to record all input changes in ETEngine, rather than just the current state of the scenario. This may be on the cards *eventually*, as it would allow us to show the full history of (saved) scenarios.

> perhaps we should rephrase the expanding text slightly to be more general: "How do updates of the ETM affect scenario's?" for instance.

Agreed. I'll change this.

ChaelKruip:
I pushed some translations

michieldenhaan:
Thanks @antw, this is indeed much better.

One thing I'm curious to hear your and @ChaelKruip's opinion on: The current text focuses on migrations and states that changes are small / affect a limited amount of scenarios. While this is true, it may leave the impression that scenario _outcomes_ are largely stable over time. This isn't necessarily the case as scenario outcomes may also change due to us improving parts of the ETM (e.g. annual heat calculation -> time resolved calculation), updating specs etc for which no migration or only a partial migration is possible. The impact of this could be quite large and affects nearly all scenarios. Do you think this is worth mentioning?

A second (more minor) point is that new sliders could have been added and their 'default' value may not necessarily be in line with the intentions of the scenario maker. E.g. the heat storage toggle will be 'off' for all pre time resolved heat scenarios, tho this does not mean that this is an active decision by the scenario maker.

antw:
Fair points. I suggest removing the paragraph about most scenarios being unaffected, make it clearer that we cannot *always* respect the intentions of the author, and that outcomes can change.

> During updates we must sometimes make small changes to the saved settings for some sliders. This is done to ensure that scenario remains compatible with the new version of the model and, whenever possible, that the intentions of the scenario author are preserved. Scenario outcomes may have changed due to improvements in the model.

If you prefer, we could also be more explicit with something like this:

Screenshot 2021-02-08 at 13 52 41

michieldenhaan:
👍 perfect! I prefer the more explicit version

antw:
@michieldenhaan Thanks for the feedback. [Some more translations are needed](https://github.com/quintel/etmodel/blob/6f567f510b29ccc9e4b9ffe226c2554202567b69/config/locales/nl_etm.yml#L193-L201), please. :pray:

michieldenhaan (approved):

PR #3607, opened 2021-02-09
Fix bug where columns of energymix were overlapping

noracato:
The columns of the EnergyMix infographic were overlapping:

Before

I fixed it to look like this:

After

(Don't look at the exact numbers, these are two different scenarios 😉 )


MartLubben:
Ziet er goed uit!

MartLubben (approved):

Issue #3622, opened 2021-02-19
Group sub headers should be changeable for each group


mabijkerk:
It is currently possible to specify both a general sub header and a group sub header. See for example the slide`Passenger transport - Applications`, where both the "%/year" growth and the "share" of each transport mode is shown:
<img width="469" alt="Screenshot 2021-02-19 at 11 02 25" src="https://user-images.githubusercontent.com/67338510/108489541-f58ea700-72a1-11eb-806f-dbc963b0c9d9.png">

For some slides it is desirable to specify a different sub header for each group. See for example the slide `Excess electricity - Storage`, where the first group `Low voltage network` should show "share", but the other two groups should show "electricity input".
<img width="460" alt="Screenshot 2021-02-19 at 11 05 17" src="https://user-images.githubusercontent.com/67338510/108489858-5c13c500-72a2-11eb-806f-f92c62c97da0.png">

Currently it does not seem possible to specify different sub headers for each group, is this something that can be added @antw? Another example of where this is an issue is the slide `District heating - Heat sources`, where all the sliders are referred to as "heat output" even though some quite clearly set a share.

antw:
This is possible, but I admit it's really not obvious how to do it.

The slide needs to have a `group_sub_header` attribute. Doing this for the storage slide (and removing `general_sub_header`) looks like so:

![Screenshot 2021-02-19 at 14 39 52](https://user-images.githubusercontent.com/4383/108518365-5f1dae00-72c0-11eb-84c3-e27051907bcf.png)

The sub-headers currently don't look so good when the text is more than one word, as in this case. You may then customise the labels for each group [by adding translations to the "subheaders" collection](https://github.com/quintel/etmodel/blob/7d12ed30bc05e62f67517cba5ec29d784b99ca59/config/locales/en_etm.yml#L140-L162). These should match the `interface_group` of the sliders.

For example, [the interface groups of the storage sliders are "lv_network", "mv_network", and "hv_network"](https://github.com/quintel/etmodel/blob/7d12ed30bc05e62f67517cba5ec29d784b99ca59/config/interface/input_elements/flexibility_flexibility_power_storage.yml). Adding translations for those sub-headers...

```yaml
lv_network: "share"
mv_network: "capacity"
hv_network: "capacity"
```

... results in:

![Screenshot 2021-02-19 at 14 37 57](https://user-images.githubusercontent.com/4383/108518764-d9e6c900-72c0-11eb-85d1-fd4ffcd341a2.png)

[...]

Issue #3619, opened 2021-02-18
Hydrogen production sliders should clearly state whether they affect input or output capacity


mabijkerk:
The sliders for hydrogen production technologies currently do not clearly show whether they set the installed hydrogen output capacity or the input capacity of the energy source used to produce the hydrogen. This can lead to confusion about what the actual **hydrogen** production capacity in a certain scenario is (see for example issue #3618).

For other technologies it is clearly stated which type of capacity is meant, for example:
<img width="300" alt="Screenshot 2021-02-18 at 10 46 58" src="https://user-images.githubusercontent.com/67338510/108338515-a29bed80-71d6-11eb-8114-7dbb921c5a00.png">  <img width="300" alt="Screenshot 2021-02-18 at 10 47 34" src="https://user-images.githubusercontent.com/67338510/108338589-b8111780-71d6-11eb-8f86-ac5330a94725.png">

For hydrogen this is currently not the case, for example:
<img width="300" alt="Screenshot 2021-02-18 at 10 49 43" src="https://user-images.githubusercontent.com/67338510/108338864-04f4ee00-71d7-11eb-8c36-c30d0d7c1e9e.png"> <img width="300" alt="Screenshot 2021-02-18 at 10 50 01" src="https://user-images.githubusercontent.com/67338510/108338908-10481980-71d7-11eb-9d81-959c1dd22d5b.png">

I propose to add a similar pointer for the hydrogen sliders to indicate whether input or output capacity is meant.

A quick scan through the model shows me that this is not consistently applied for other technologies either. For example, biomass plants also do not show whether input or output capacity is meant. In some cases it could be considered evident whether input or output capacity is meant. @michieldenhaan do you think these pointers should be added for other technologies as well?

michieldenhaan:
Yes! Preferably every MW slider should indicate whether it is input or output capacity.

marliekeverweij:
Adding this to the March Deploy milestone. Do you have time to take care of this @mabijkerk ?

PR #3626, opened 2021-02-22
Subheaders update for electricity storage and district heating supply

mabijkerk:
As mentioned in issue #3619 it was not clear for some hydrogen input sliders what type of input the sliders actually set. It turned out that this was the case for electricity storage and district heating supply as well.

- For electricity storage no subheader was specified at all because the value of the slider actually differed for each interface group. In #3622 it is explained that each interface group **can** have a different subheader.
- For district heating supply a single subheader "heat output" was specified for all sliders, even though this subheader did not apply to some of the sliders (for example the carbon emissions of imported heat).

For both electricity storage and district heating supply subheaders specific to each interface group have been added.


marliekeverweij:
Looks gooooood!

marliekeverweij (approved):

Issue #3636, opened 2021-02-24
Improvements to scenario reports


marliekeverweij:
@mabijkerk thoroughly tested the scenario reports and in this issue you can find the quick fixes that would be nice to pick up before the deploy:

**All reports:**
- Remove texts and charts about biofootprint -> these are outdated. We haven't updated the bio-footprint calculation recently, so the chart shows incorrect numbers.

**National report (English):**
- Issue: At a glance.. > More information: "In your scenario, primary energy use has been reduced to 67.4% of its value in 2015." when I click on "Learn more", nothing happens. 
This link should direct to: /scenario/reports/main#energy-use-and-production <img width="658" alt="Screenshot 2021-02-22 at 16 17 44" src="https://user-images.githubusercontent.com/32833996/108990104-b348d980-7696-11eb-9f6b-3b59a04ea498.png">

- "Profitability of power plants" should be renamed to "Profitability of dispatchable power plants" as it only considers dispatchable power plants and not must-run / volatiles (like wind turbines)

**National report (Dutch):**
- Samenvatting scenario voor Nederland: "De totale huidige uitstoot van het energiesysteem is nu 161.47 MT en 83.9% van de totale broeikasgasuitstoot van regio." This sentence seems incorrect,"regio" should be replaced with the actual region in the scenario, in this case "Nederland".
- Samenvatting scenario voor Nederland: "Het Energietransitiemodel richt zich enkel op het energiesysteem van Nederland." & "Hieronder is te zien hoe de energetische CO2-uitstoot zich ontwikkelt in het scenario." This has to be updated because we now do show some of the non-energetic emissions (which the corresponding chart also shows).

**Regional scenario report (Dutch):**
- Similar to the Dutch national report, the text still states that only energetic emissions are taken into account. This should be updated.
- When you open a blank regional scenario and don't change the energy consumption, it is stated that the energy consumption is 0% of the initial energy consumption, while I think that it should be 100% or we could say that there is 0% change... You can decide :)  See for example a blank scenario for Ede: <img width="725" alt="Screenshot 2021-02-22 at 17 18 28" src="https://user-images.githubusercontent.com/32833996/108990837-aaa4d300-7697-11eb-9a77-66a0ae13e019.png">

[...]

PR #3639, opened 2021-02-24
Scenario report: removal biofootprint data and minor changes, also includes a small textual change in appliances, solves #3636

Charlottevm:


marliekeverweij (changes requested):
Looks really good!

Two comments:
- In the Dutch national report the two Renewability-charts have been removed as well. I think these could stay :) See: https://github.com/quintel/etmodel/pull/3639/files#diff-daebe6c5084af2877bd9e8077cbda9a18b6d4ae3f1d1bc9cc80f3e8ead4ae23eL568-L581
- I see that the issue that you've solved for the Dutch regional report is also present in the national report, could you make the same adjustments there?
Screenshot 2021-02-24 at 16 20 57

marliekeverweij (approved):

PR #3646, opened 2021-03-01
Fix uneven letter spacing in Firefox chart PNGs

### LINKED ISSUE #3630: Increase spacing in chart-as-image download
EDIT: this does not happen in Google Chrome or Safari, just in Firefox

In a few chart-as-image downloads I came upon the following spacing issues:

Transport fuel mix undefined

Not deploy critical in my opinion, but would be nice to fix whenever one of us has some time left.

antw:
The letter spacing is uneven when using the system font in Firefox on a HiDPI screen. For that (rather rare) combination, fall back to Helvetica / Arial which doesn't have the problem.

Letter spacing issues seem to be somewhat common with html2canvas and there are several issues already open, so I don't think I'll send this upstream to them.

Closes #3630

### DIFF
--- a/app/assets/stylesheets/_chart.sass
+++ b/app/assets/stylesheets/_chart.sass
@@ -101,6 +101,10 @@ $chart-header-icon-color: #8895b2
     top: -10000px
     left: -10000px
     position: absolute
+    &.firefox-hidpi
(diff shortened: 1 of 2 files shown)


noracato (approved):
Wow, that's an awfully niche case... Thanks for fixing it!

PR #3648, opened 2021-03-03
Add 2021 ETM survey

### LINKED ISSUE #3612: Add a short survey
As part of the UX improvement project, we'd like to add a short, *optional* survey which ETM visitors may fill out. There are four or five questions we'd like to ask in order to get a better idea of who uses the ETM, and for what purpose.

The proposal is for a small pop-up to appear in the lower-right of the screen asking the first question. Perhaps something like "Please rate your experience with energy transition" on a scale of 1-5:

IMG_36266FB528ED-1

* If the visitor answers the question, we store the response in the database and show the next question.
* At any time, they can close the pop-up and not be bothered by it again.

The exact inputs for the survey are not yet known and will depend on the questions created by @Charlottevm. It's likely some will be "choose a value", for example, from 1 to 5, while others may be multiple choice. I think we should shy away from text inputs since I suspect (albeit without evidence) that people are less willing to answer these.

The admin section should have a way of viewing the responses, or perhaps just the ability to export them all as a CSV.

@ChaelKruip I'd like to propose this for the next dev sprint. It will hopefully take only a couple of days, and we might be able to sneak it in before the feature freeze.

antw:
This PR adds a five question survey to the scenario page. This is part of the project to improve the user experience, and also helps the front-end refresh.

* The survey appears only on the scenario page, in the lower-right corner (where's there's normally blank space). We played with including it on the front page, but decided against doing so as some visitors will be completely new to the ETM; throwing a survey at them isn't the best introduction.

* Answers are sent as soon as the visitor makes a choice, rather than sending them all at once when finished. This means if a visitor abandons the survey part-way through, we still (hopefully) get something useful. They may also later resume where they left off.

* **Ahhh! Don't look at all the jQuery!** React can't come to the ETM soon enough. A lot of CSS and JS was needed to animate everything nicely, but I think it was worth it. It's only temporary, anyway.

* The popup is somewhat obnoxiously animated because we *really* want the visitor's attention. If they continue to ignore it, it'll shake at them every 30 seconds until they either proceed with the questions or choose to dismiss it. This animation will stop whenever the visitor mouses-over the survey, or proceeds to the first question.

* If you dismiss the survey by pressing "x" or "Ask me later", it'll go away for three days. If you choose "Never ask again", it'll go away forever. ~~You can undo the dismissal by clearing your local storage~~ (Edit: You can reopen the survey from "Actions" on the grey scenario nav bar). If you've already completed the survey you must also sign in and out of the ETM to clear your session if you want to try it again.

* Survey results are stored in the DB and can be downloaded as a CSV in the admin section. The CSV includes user IDs but does not contain any personally identifiable information.

* SVG icons were either MIT licensed or free with attribution. I've added credits to our README until we find somewhere better.

Screenshot 2021-03-03 at 14 46 09

Screenshot 2021-03-03 at 14 46 50

Screenshot 2021-03-03 at 14 48 12

Closes #3612


noracato (approved):
What a fun survey! I loved the emojis and the confetti at the end!

The edit that made it able to reopen the survey is a nice addition. 

I fixed some typos in the Dutch translations of the survey. Otherwise it looks great! (ofcourse I'm not going to say anything about the jQuery 😉, this is becoming a nice pre-refresh running joke)

PR #3652, opened 2021-03-08
Add sign up link to the main header

antw:
This PR adds a "Sign in" link to the ETM header.

A while back we redesigned the main ETM header; adding feedback and language options, a sign-in link when visiting as a guest, and links to your account and scenarios when you're a registered user.

At that time, the drop-down sign in form was moved to a separate page, and with it went the "Sign up" links. Visitors now need to visit the "Sign in" page in order to create an account.

I didn't think this was a huge issue and still don't have any data to suggest that it is – there are plenty of people creating accounts – but I'm having second thoughts. I'm suggesting changing this now to make it clearer to new visitors how to create an account; some may not realise that clicking "Sign in" also provides instructions for signing up.

This makes the header busier, but I'm leaning towards thinking it's a worthwhile trade-off.

@ChaelKruip @noracato I realise this is a tiny change and hardly worthy of a PR, but I'd appreciate a second opinion.

Screenshot 2021-03-08 at 13 48 16

Screenshot 2021-03-08 at 13 53 09


ChaelKruip (approved):
I think this is a good idea! Perhaps we can make the two options more distinct by calling them "login" / "inloggen" and "sign up" / "registreren"?

noracato (approved):
I like it as well, thank you for asking. Even without the front-end refresh having truely started the ETM is getting nicer every week :)

Issue #3651, opened 2021-03-08
Merit order table should show "Biogas CHP" instead of "must run CHP"


marliekeverweij:
At the moment the merit order table has the row "must run CHP" includes the node group `must_run_electricity_production` and this group includes the biogas CHP node and waste power node. See [this query](https://github.com/quintel/etsource/blob/master/gqueries/output_elements/output_series/table_116_merit_order/merit_order_must_run_capacity_in_merit_order_table.gql)
![Screenshot 2021-03-08 at 14 36 32](https://user-images.githubusercontent.com/32833996/110328467-b12e3580-801b-11eb-97b3-21f35125888e.png)

However, the waste power is already separately shown in the merit order table, so it's now double!

So instead of the "must run CHP" we should separately show only the "biogas CHP"

![Screenshot 2021-03-08 at 14 41 56](https://user-images.githubusercontent.com/32833996/110329097-77116380-801c-11eb-934c-19418472713e.png)

ChaelKruip:
Good catch @marliekeverweij !

mabijkerk:
The issue addressed above for the merit order table is fixed by https://github.com/quintel/etsource/pull/2439 and https://github.com/quintel/etmodel/pull/3683.

However, in the merit order chart both the Biogas CHP and Waste incinerator are still represented by the combined group "Must-run CHP". This is misleading, given that the Waste incinerator is not a CHP. To align the merit order chart with the merit order table, the "Must-run CHP" group in the merit order chart should be split into the two underlying nodes. This issue has been renamed accordingly.

Another minor issue that should be tackled is the consistency in labels between the merit order chart and the merit order table. For example, in the merit order table solar PV plants are described as "Solar PV (ground-mount)", while the merit order charts uses "Solar PV plants".

PR #3683, opened 2021-04-02
Add local CHP biogas to merit order table and remove must-run CHP

mabijkerk:
Addresses issue https://github.com/quintel/etmodel/issues/3651 by adding local CHP biogas to the merit order table and removing must-run CHP. This pull request also includes minor updates to labels used in the table.


marliekeverweij (approved):
Super!

PR #3729, opened 2021-05-19
Clarify investment costs in node details

### LINKED ISSUE #3727: Misleading "initial investment" label in technical properties table
A visitor to the ETM recently sent a message expressing confusion why after reducing solar PV investment costs by -100% the technical properties table still shows 500k initial investment.

You may reproduce this yourself by visiting Costs → Solar power → Investment, setting the Solar panels slider to -100%, then viewing the technical properties.

Screenshot 2021-05-18 at 12 56 26

This happens because what we are calling "Initial investment" is actually `total_investment_over_lifetime_per(:mw)`. There are a number of other places where we are mislabelling *total* investment as *initial* investment.

Total investment is defined as:

* total_investment_over_lifetime
  * total_initial_investment
    * initial_investment
    * ccs_investment
    * cost_of_installing
    * storage_costs
  * decommissioning_costs

The 500k costs come from `cost_of_installing`; this *is* communicated in the description of the investment slider but is not obvious in the table itself.

There are two solutions:

1. Relabel the row as "Investment over lifetime"
2. Remove the initial investment row, and replace it with five rows – one for each contributor in total investment. This would be clearer to users, but risks the table becoming outdated if we change how total investment is calculated (if we add another contributor, we must remember to also update this table).

I'm happy to take care of this (but would need some assistance with Dutch translations).

@marliekeverweij @michieldenhaan: Since you both had some involvement in updating this table recently, I'd appreciate your feedback. Including @ChaelKruip.

antw:
Makes a few small changes to the technical and financial properties table:

* Rephrases "Initial investment" as "Investment over lifetime" when using the `total_investment_over_lifetime` node attribute, to clarify that the row contains not just the initial investment, but also other costs such as installation and decommissioning.

* Adds a toggleable description of the investment costs. Descriptions may be added to any attribute using the "node_attributes.descriptions" locale namespace.

A couple of translations are needed, please and thank you:

* ["Investment over lifetime" title](https://github.com/quintel/etmodel/blob/434333e22dc0ec9ffe8d2a0f12bfcf7c47e88873/config/locales/nl_node_details.yml#L45-L47) (repeated 3 times)
* [Attribute description](https://github.com/quintel/etmodel/blob/434333e22dc0ec9ffe8d2a0f12bfcf7c47e88873/config/locales/nl_node_details.yml#L59-L67) (repeated 3 times)

Closes #3727

### DIFF
--- a/config/locales/en_node_details.yml
+++ b/config/locales/en_node_details.yml
@@ -57,13 +57,13 @@ en:
       wood_pellets_input_conversion: Wood pellets input conversion
     descriptions:
       total_investment_over_lifetime_per_mw_electricity: |
-        Includes initial invstment with any applicable installation, CCS, storage, and
+        Includes initial investment with any applicable installation, CCS, storage, and
         decommissioning costs.
       total_investment_over_lifetime_per_mw_typical_input_capacity: |
-        Includes initial invstment with any applicable installation, CCS, storage, and
+        Includes initial investment with any applicable installation, CCS, storage, and
         decommissioning costs.
       total_investment_over_lifetime_per_plant: |
-        Includes initial invstment with any applicable installation, CCS, storage, and
+        Includes initial investment with any applicable installation, CCS, storage, and
         decommissioning costs.
     download_analysis: Download source data
     groups:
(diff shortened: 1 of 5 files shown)


config/locales/en_node_details.yml
```diff
@@ -57,13 +57,13 @@ en:
       wood_pellets_input_conversion: Wood pellets input conversion
     descriptions:
       total_investment_over_lifetime_per_mw_electricity: |
-        Includes initial invstment with any applicable installation, CCS, storage, and
+        Includes initial investment with any applicable installation, CCS, storage, and
```
antw, line 60: TY! 👍

marliekeverweij (approved):

PR #3748, opened 2021-06-28
Added solar thermal collective to ovm-csv NOT TESTED

MartLubben:
Solar thermal collective heat was missing in the OverMorgen datadump csv.
Maybe ask @Charlottevm to test it.


MartLubben:
It works in combination with: https://github.com/quintel/etsource/pull/2479

Charlottevm:
@antw Do you have time to shortly check this before the deploy? We do not understand why Semaphoreci is giving a failing check..

antw:
It's fixed now.

marliekeverweij:
Great @antw! Was this something we could fix next time?

antw:
> Was this something we could fix next time?

Some of ETModel's tests fetch data from ETEngine. This can be problematic as it places extra load on ETEngine and, more importantly, means that the tests aren't always reproducible: different data might be used each time the test is executed.

We use a library called VCR which saves the results of these requests to YAML files, with the data in these files use during future test runs, instead of making new requests to ETEngine. If VCR detects that you're requesting different data than was used to generate the YAML files (as was the case in this request: an extra query was added) then it will raise an error letting you know which file is to blame:

```
================================================================================
An HTTP request has been made that VCR does not know how to handle:
  PUT https://beta-engine.energytransitionmodel.com/api/v3/scenarios/648695
  Body: [truncated ...]

VCR is currently using the following cassette:
  - etmodel/spec/cassettes/SavedScenarioReportsController/GET_show/ovm_csv/has_a_report_template.yml
    - :record => :once
    - :match_requests_on => [:uri, :method, :body]

[truncated ...]
================================================================================
```

The key line in the message is:

```
VCR is currently using the following cassette:
  - etmodel/spec/cassettes/SavedScenarioReportsController/GET_show/ovm_csv/has_a_report_template.yml
```

This provides you with the path to the responsible YAML file. **Delete this file, then re-run the tests locally (`bundle exec rspec`) to generate fresh versions, then commit the changes.**

The tests for this PR had three failures, each one due to a different YAML file. I deleted all three, ran the tests, and committed the new YAML files.

marliekeverweij:
Great Anthony! Thanks for the clear explanation :) Good to know

Issue #3755, opened 2021-07-07
The FLH mentioned under 'Default' for weather years are not in line with actual FLH for NL2019


AlexanderWirtz:
![image](https://user-images.githubusercontent.com/308563/124785205-f4a39c80-df46-11eb-823e-c75627802dde.png)

Do we even need to mention these here?

re-assign as needed

PR #3765, opened 2021-07-16
Adjust text for default weather year

marliekeverweij:
The current text doesn't match with the nl2019 dataset when that comes online. I thought it would be nice to add a link to the 'Modify profiles' section, so users can have a look at the current profiles. However, it is not supported at the moment. @antw / @noracato is that easy to adjust?

Screenshot 2021-07-16 at 11 03 44

Closes https://github.com/quintel/etmodel/issues/3755


config/locales/en_radio_elements.yml
```diff
@@ -5,11 +5,9 @@ en:
...
         title: Default
         description: |
-          The wind and solar curves for the start year
-          are based on actual production curves of ENTSO-e. The temperature
-          curve is based on KNMI-data.
-          (Full load hours: wind onshore = 1920, wind coastal = 2550, wind offshore = 3500,
-          solar = 867)
+          By default the ETM uses production curves from the start year of your scenario.
+          The corresponding full load hours are shown on this page. You can have a look at the
+          production curves in the <a href='https://pro.energytransitionmodel.com/scenario/flexibility/curve_upload/upload-curves'>Modify profiles</a> section.
```
redekok, line 10: It might be useful to use a relative hyperlink here--i.e., a hyperlink in which the hard-coded "pro.energytransitionmodel.com/..." isn't included. Now, when testing this PR locally, the hyperlink directs me to a different scenario because it points at pro instead of my localhost.

redekok (changes requested):
LGTM! Just added a minor suggestion regarding the hyperlink to the profiles page. 

Furthermore, just out of curiosity: why did we rephrase this part? To make it more generic in case the FLH values change for a dataset?

config/locales/en_radio_elements.yml
marliekeverweij: Good point, I'll adjust it

marliekeverweij:
Thanks for reviewing :) I've pushed a fix for the issue

>Furthermore, just out of curiosity: why did we rephrase this part? To make it more generic in case the FLH values change for a dataset?

Yes for `nl2019` (and maybe already for some local datasets??) the FLH will be different in the default dataset

redekok (approved):

PR #3770, opened 2021-07-22
Add hydrogen combi heater

### LINKED ISSUE #3759: Add sliders for hydrogen combi boiler in households and buildings
The ETM currently only allows hydrogen for heating to be used in hybrid heat pumps in the domestic and services sector; there is no combi boiler heater that uses 100% hydrogen.

noracato:
Add inputs and alter relevant charts for hydrogen combi heaters.

Closes #3759


marliekeverweij:
Looks goooooood @noracato !

Two minor remarks:
- label is missing Screenshot 2021-07-23 at 15 52 57
- Variable costs return `null` and it should say zero
Screenshot 2021-07-23 at 15 13 05

noracato:
Fixed the translation in the chart, but the other issue you had I cannot reproduce. When I look at the properties table (in the household section) myself, I get this:

Screenshot 2021-07-26 at 09 45 00

Where the initial investments are 0 (as they should be) instead of NULL...

marliekeverweij (approved):

Issue #3749, opened 2021-07-02
Adding future efficiencies to the Costs section


Charlottevm:
@quintel/core:

As part of the NI updates we want to include the future efficiencies of technologies within the Costs section and possibly restructure the cost section. If you have an opinion on the matter then we would love to hear what you think about our suggestions below. 

The update regards the following technologies: 

**Electricity**

_Coal plants_
- Pulverized coal 
- Pulverized coal with co-firing 
- Coal IGCC 
- Coal conventional 
- Coal plant for district heat (CHP) 
- Co-firing coal plant for district heat (CHP) 

_Gas plants_
- Gas CCGT 
- Gas turbine 
- Gas conventional 
- Gas combustion engine 
- Large-scale gas plant for district heat (CHP) 
- Small-scale gas plant for district heat (CHP) 

_Oil plants_
- Oil-fired
- Diesel generator

_Nuclear_
- Nuclear conventional
- Nuclear 3rd Gen

**Renewable electricity**
- Solar
- Hydro
- Biomass
- Waste power
- Hydrogen plant
- Geothermal

**Heating**

_Households_
- Condensing combi boiler
- Heat pump (ground)
- Wood pellet stove
- Electric heater
- Gas-fired heater

_Buildings_

- Gas-fired heater
- Heatpump with TS
- Electric heater
- Biomass-fired heater

We have two options at the moment:

**1. We simply add the efficiencie sliders to the existing technologies in the cost section like so:**
<img width="473" alt="afbeelding" src="https://user-images.githubusercontent.com/74194356/124261474-207ee680-db31-11eb-920f-8ef203b09372.png">

**2. We add the efficiency but in addition we also restructure the sidebar of the cost section as follows:**

**Costs**
- Merit Order
- Fuel Prices
- Electricity
- Renewable electricity
- Imported electricity 
- Infrastructure
- Heat
- Transport
- Hydrogen
- Flexibility
- CCUS
- WACC

_We are currently in favor of option 2 as it would improve the user experience due to the recognition of the sidebar items compared to the suppy section. Which option do you prefer? & do you think this is the right (most intuitive) place to add the future efficiencies for technologies?_

Furthermore we were wandering whether the sliders describing the 'investment' as well as 'operation & maintenance' should be combined into one slider (describing the LCOE?). This would reduce the number of sliders and possibly make it easier for users to use the cost section. 

_Is this an important asset of the cost section? What are possible hurdles in doing this? Do you agree this would be easier for users?_

[...]

Issue #3016, opened 2019-04-08
Shall we narrow down the number of sidebar items in the Costs-section?


marliekeverweij:
Currently the 'Costs'-section has 14 categories:

![Screen Shot 2019-04-08 at 17 15 20](https://user-images.githubusercontent.com/32833996/55735905-8ea3d400-5a22-11e9-9785-ac7154726cde.png)

Shall we narrow it down by creating new sidebar items and nesting current categories inside them?

For example, we could create the following sidebar items:
- 'Power plants' : which could include 'Combustion plants' and 'Nuclear plants'
- 'Renewable electricity': which could include 'Wind tubrines', 'Solar power', 'Geothermal' and 'Hydro electric'

Let me know what you think and if you have better suggestions! @AlexanderWirtz, @antw  , @ChaelKruip , @DorinevanderVlies , @jorisberkhout , @redekok , @MartLubben , @michieldenhaan

DorinevanderVlies:
I like this plan!

Furthermore I think we can remove the slide "merit order positions" as it has overlap with the slide "Merit order price" in "Results" --> "Data export"

PR #3771, opened 2021-07-22
Adding future efficiency sliders to "Costs & parameters" section

marliekeverweij:
Thanks to @Charlottevm we now have future efficiency sliders for power plants and heating technologies in the built environment! Closes https://github.com/quintel/etmodel/issues/3749
See issue for details about the eaxt sliders that have been added.

Screenshot 2021-07-22 at 17 07 57

Besides adding the new sliders, we picked up the following issues:
- Restructured and renamed the Costs section, it's now called "Costs & parameters"
Closes https://github.com/quintel/etmodel/issues/3016
- The merit order page has been moved to "Supply -> Electricity" as it better fits there (heat merit order is also located under Supply)
- [The transport efficiency sliders](https://pro.energytransitionmodel.com/scenario/demand/transport/efficiency-improvement) have been moved to "Costs & parameters" as these sliders change the efficiency of transport vehicles. We haven't moved the [efficiency sliders of industry,](https://pro.energytransitionmodel.com/scenario/demand/industry/chemicals) as these sliders are not actually changing the efficiency of a technology, but instead decreasing the electricity / heat demand.
- We have added a slider to adjust the efficiency of power-to-gas elektrolyzers
- Updated some costs sliders (many were missing technologies or had to be split up with the new structure)


jorisberkhout:
Very nice work @marliekeverweij and @Charlottevm ! I've tested the new feature extensively and, as discussed with @marliekeverweij , I'm reporting my findings below. I suggest resolving these before merging:

- [ ] Position of the merit order (sub) sidebar item. I had a hard time finding it because it's hidden unless you open the Supply ->  Electricity sidebar item. I suggest promoting it to a real sidebar item, preferably right after the Renewable Electricity sidebar item
- [ ] I find the name of the Costs & parameters tab a little vague and would suggest changing it to Costs & efficiencies
- [ ] The ranges of the costs sliders imply that you can make it possible to build power plants for free. I know that this is the users own responsibility, so action needed, but I just wanted to note this
- [ ] Definition of the costs sliders. In the industry demand part of the ETM we use similar looking sliders to set the size of industrial subsector. Here, the sliders are defined such that 100% means that the sector stays the same size it presently is. The costs sliders are defined such that 0% means that the costs stay the same. I suggest no changes to any of the sliders, but to keep this in mind for future development projects.
- [ ] I find the heat efficiency sliders for the CHPs a little oddly positioned in the electricity sidebar items but have no suggestions on how to improve this
- [ ] In the Total cost of electricity production chart, the numbers of the Y-axis overlap with the axis itself if the total costs of production go above 1000 EUR/MWhe
- [ ] In the Total cost of electricity production chart, the gas CHP is called Central gas CHP, whereas in the sliders they are called Large-scale gas CHP. I suggest renaming them in the chart to Large-scale gas CHP
- [ ] I'm missing the O&M costs for solar power. Is this deliberate?
- [ ] The COP for the Heat pump (ground) is 4.8 according to the efficiency slider. However if I go to the Technical and financial specifications of the related demand slider, it says the COP is 5. Is this a rounding error in the Technical and financial specifications?
- [ ] In the costs -> heat slide, it struck me that the Households gas-fired heater has an efficiency of 90% whereas the Buildings equivalent has an efficiency of 107%. I suggest renaming the latter (and do the same for the demand sliders that it corresponds to).
- [ ] In the CCUS slide of the costs tab, all O&M sliders are called Onderhoud en Beheer in Dutch, whereas all other O&M slider are just called Onderhoudskosten. I suggest changing the Onderhoud en Beheer to Onderhoudskosten to be in line with all other sliders.
- [ ] In a scenario for a local dataset (I tested Beverwijk), the COP slider for the Heat pump (ground) is scaled (in this case to 0.0218). This should not be the case.
- [ ] The Dutch translation for the Kosten & Parameters -> Warmte -> Other heat sources is missing.

If you have any questions about the list above, let me know!

marliekeverweij:
Thanks for your feedback :) 

I have tackled most of the issues with an action in my previous commit.

> In the Total cost of electricity production chart, the numbers of the Y-axis overlap with the axis itself if the total costs of production go above 1000 EUR/MWhe

I cannot reproduce this issue. @jorisberkhout could you create a separate issue for this?

>  I'm missing the O&M costs for solar power. Is this deliberate?

No this is not dileberate, but we focussed on the efficiency sliders. I've added it to the costs slider improvements issue.

>  The COP for the Heat pump (ground) is 4.8 according to the efficiency slider. However if I go to the Technical and financial specifications of the related demand slider, it says the COP is 5. Is this a rounding error in the Technical and financial specifications?

I will create an issue for this

Then I think we are good to go, right? :)

jorisberkhout:
>> In the Total cost of electricity production chart, the numbers of the Y-axis overlap with the axis itself if the total costs of production go above 1000 EUR/MWhe
>
> I cannot reproduce this issue. @jorisberkhout could you create a separate issue for this?

Will do!

> Then I think we are good to go, right? :)

Yes, nice work! I'll approve this PR!

jorisberkhout (approved):

PR #3776, opened 2021-07-27
Add oil mix sliders and chart

### LINKED ISSUE #3760: Make it possible to set oil mix for all sectors
@quintel/core if you would like to think along with this improvement for Northern-Ireland, you are welcome to do so!

The ETM currently does not allow to set the mix of oil in a sector: all final demand oil is always `crude_oil`. 

This results in two short-comings:

1. It is currently not possible to use **bio-fuels** in oil-fired heaters
2. It is currently not possible to use a **different type of oil** in oil-fired heaters. Using a different type of oil (for example using kerosene instead of diesel) could result in lower CO2 emissions and we cannot model this at the moment.

Oil-fired heaters are common in many countries, therefore we want to improve the two points above.

We want to improve this by adding a oil mix slider group for each sector for _energetic oil_. In this slider group (adding up to 100%) the user could for example set the oil mix, choosing from:
1. kerosene
2. bio-kerosene
3. diesel
4. bio-diesel
5. other oil (we use `crude_oil` in the back-end)
6. other bio-oil

This slider group will be added for the following sectors: households, buildings, industry, agriculture. So in total 4 slider groups will be added.

**Questions to you:**
- Are there more fuels you would like to specify?
- Shall we aggregate all bio-fuels into one general 'bio-oil' slider? This brings the number of sliders down to 4. EDIT: we will not aggregate and keep these six fuels.
- What is a good location for these sliders in the front-end? The Supply section feels right. I do not want to add a new side bar item to the supply section. I think the Supply -> Biomass-section fits best, as it's also about the bio fuels used. We can link from the oil-fired heater slider texts to this section.

@michieldenhaan suggested that for industry it might be better to specify the oil-mix for each sub-sector, as the sectors are so different from each other. I would propose to do this at a later stage when we better know if this is desired. For know we will stick with one slider group for industry.

Back-end:
We want to keep the `final_demand_crude_oil`-nodes. Keeping this node means that we do not have to make a lot of adjustments to final demand charts and calculation. In the charts you will see 'oil' and not the specific types. 
Right from this final_demad-node we will add a mixer (just as in the transport sector) where the different fuels come together. The different fuels all have their own node right from the mixer. In these nodes the energy balance values are directly imported.

noracato:
Adds the possibility to set the oil mix for (mainly) heating purposes per sector.
A new slide is added to the Supply > Biomass section:

Screenshot 2021-07-27 at 16 58 15

To accompany this slide, there is a new mekko showing the total oil mix for the scenario. The mekko will also contain non-energetic use oil, and oil use in sectors that are not on the new slide, as it describes the full supply and demand of oil in the scenario.

Screenshot 2021-07-27 at 17 01 21

If you think the colours are too funky, you can change them 🌈 

---

See quintel/etsource#2496
Closes #3760


config/interface/output_elements/supply_biomass.yml
```diff
@@ -85,3 +85,18 @@
   dependent_on:
   description:
   output_element_type_name: category_bar
+- under_construction: false
+  unit: PJ
```
redekok, line 89: The gqueries report values in MJ whereas the mekko expects values in PJ. This mismatch results in gigantic values in the model (as can be seen in the chart below). 🤓 

image

config/locales/interface/input_elements/en_supply_biomass.yml
```diff
@@ -94,11 +94,179 @@ en:
...
+        Percentage of kerosene in oil products used in the domestic sector. This kerosene is mainly used
+        for heating and hot water.
+    households_bio_kerosene_in_crude_oil_share:
+      title: Bio kerosene
+      short_description:
+      description: |
+        Percentage of bio kerosene in oil products used in the domestic sector. This bio kerosene is
+        mainly used for heating and hot water.
+    households_lpg_in_crude_oil_share:
+      title: Lpg
```
redekok, line 130: Nitpick: in other parts of the model we use all caps ("LPG" instead of "Lpg"). I would be in favour of using all caps here as well.

redekok:
@noracato: Perhaps we should hide the agriculture oil mix sliders when there is no agricultural demand in a dataset (e.g. Germany)? I guess we would need a `has_agriculture` dependency for that?

redekok (approved):
LGTM!

Issue #3766, opened 2021-07-16
Slider text air source heat pump should be improved


marliekeverweij:
We recently received quite a few questions about our COP calculation for heat pumps. The slider text does not explain that the COP is calculated on an hourly basis based on the outside temperature curve. Additionally, it would be good to link to our documentation page about heat pumps: https://docs.energytransitionmodel.com/main/heat-pumps

![Screenshot 2021-07-16 at 16 38 41](https://user-images.githubusercontent.com/32833996/125964950-628967de-ff01-4e62-a630-48532229e1ca.png)

Feel free to reassign :)

marliekeverweij:
Additiionally. the COP in the node properties table should be renamed to Seasonal Coefficient of Performance. See comment in this issue: https://github.com/quintel/etengine/issues/1185

PR #3782, opened 2021-07-30
Adjusting the slider info for air heat pumps (heating and cooling)

Charlottevm:
Solves  #3766

### DIFF
--- a/config/locales/interface/input_elements/en_demand_households.yml
+++ b/config/locales/interface/input_elements/en_demand_households.yml
@@ -187,21 +187,17 @@ en:
     households_cooling_heatpump_air_water_electricity_share:
       title: Heat pump (air)
       short_description: ''
-      description: "A heat pump moves heat from one place to another. Buildings can
-        use them for heating and cooling, but they are also used in refrigerators.
-        Currently, their use in newly constructed houses is becoming quite popular,
-        since European legislation dictates that domestic energy consumption must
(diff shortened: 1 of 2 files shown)


mabijkerk:
Made some minor changes. Other than that it looks good!
Two questions: first, what is the reason that the COP for cooling is fixed, while it isn't for heating? I would expect the COP for cooling to also be dependent on the outside temperature (though I am not that knowledgeable about heat pumps). Second, does the mean that the information that was included in the slider text before (about the percentage of households that can be fitted with a heat pump) is no longer relevant?

mabijkerk (approved):

Charlottevm:
Good question, the COP for cooling is indeed also flexible / dependent on the outside temperature, however, we just did not model it this way with cooling. Would be nice to add in the future but it is not as important at the moment as the cooling demand in the Netherlands is just a fraction of the heating demand (plus the capacity of installed air heat pumps is nothing near mentionable) and usually cooling only happens at high temperatures whilest heating happens under a large array of different outside temperatures. For buildings this might be interesting as the cooling demand is much higher and constantly needed throughout the year but we might need budget for this.. 

About your other question (also a good point), I removed this as it is a bit speculative. I agree with the note but I know in practice this is not the case. This is maybe also part of a wider discussion: how far do we want to go in giving suggestions for slider settings? I am not sure yet but I will write an issue discussing the general framework for writing slider information (or any other info for that matter) as I feel the slider information is often quite divergent.

Issue #3823, opened 2021-09-06
AVI CCS is missing in heat demand and supply chart


MartLubben:
![image](https://user-images.githubusercontent.com/33415156/132175780-b579e430-82ec-4fca-9afe-d041dc2370f1.png)

Notifying @mabijkerk

Issue #3817, opened 2021-08-31
CCUS for waste incineration should be added to the investment table


redekok:
The investment table now only includes the waste incinerator and waste CHP without CCS:

![image](https://user-images.githubusercontent.com/5946498/131513289-c52dbcdb-3d76-4e84-8a4c-668014aa2265.png)

MartLubben:
We are maybe removing this investment table: https://github.com/quintel/etmodel/issues/3781

Let's remove deploy September 2021. Don't you think @redekok Roos?

redekok:
Do you know when the investment table will be removed @MartLubben? It's just a minor effort to update it so I wouldn't mind fixing that before the deploy. 

 What are your thought on this @mabijkerk ?

mabijkerk:
As I understood from Mart it is not yet fully clear what will happen with the investment table. It might be decided that it will stay in the model. If @MartLubben agrees and you have time to pick this up @redekok then that would be great!

redekok:
Unfortunately by now I don't have any time to pick this up anymore anytime soon. Perhaps @Charlottevm or @MartLubben could help out here?

PR #3825, opened 2021-09-06
Added AVI CCUS to graph Demand and Supply of district heating

lottevanvlimmeren:


lottevanvlimmeren:
Closes #3823

lottevanvlimmeren:
Closes #3817

MartLubben:
It seems to be working fine. 
I only see that the Waste incinerator CCS (must-run) (node: energy_power_supercritical_ccs_waste_mix_co2) is missing from the investment table. @lottevanvlimmeren Could you add this tomorrow morning?

mabijkerk:
Before merging I would suggest also looking into #3828.

MartLubben (approved):

PR #3846, opened 2021-09-23
Add flags for NL2019

michieldenhaan:
@antw I tried adding a flag for the NL2019 dataset to the landing page, top left corner of the model and scenario report. Could you double-check whether my changes are correct?

One thing I wasn't sure about: The flag on the landing page only showed up after adding a `nl2019.png` file, even though I changed the `_root.sass` to `nl.png`
```
      &.nl2019
        background-image: asset-url("flags-24/nl.png") 
```


antw (approved):
> Could you double-check whether my changes are correct?

Looks good to me.

> One thing I wasn't sure about: The flag on the landing page only showed up after adding a nl2019.png file

Yes, I think the drop-down relies on an image file matching the dataset key.

Issue #3842, opened 2021-09-21
Chart Total cost of electricity production fails due to outdated node


mabijkerk:
When I try to open the chart `Total cost of electricity production` ETEngine produces an error for the query `investment_for_biomass_chp_in_overview_costs_of_electricity_production`. It turns out that this query refers to an outdated CHP in the agriculture sector: `agriculture_chp_supercritical_wood_pellets`.

The queries for this outdated CHP should be updated to refer to `energy_chp_local_wood_pellets`.
@lottevanvlimmeren do you have time to pick this up?

Additionally, it does not show up in the chart list, you can see it from `Costs & efficiencies > Fuel prices`. Can you add it to the list too?

mabijkerk:
It is not very clear what model changes have caused this error to pop up now, because on PRO the chart still works. As a user you can see the node but not select it:

<img width="500" alt="Screenshot 2021-09-21 at 17 22 04" src="https://user-images.githubusercontent.com/67338510/134199372-ec8f9d98-5789-4f9d-a9eb-85322b57e148.png">

@antw, as far as @lottevanvlimmeren and I can determine, the chart has not been changed on ETModel for quite some time. Do you know of any other model changes that could have suddenly caused this error? If we can determine the cause, maybe we can prevent other issues as well.

antw:
> As a user you can see the node but not select it:

If the query for the investment costs returns 0, the power plant will not be selectable.

> Do you know of any other model changes that could have suddenly caused this error?

[`V` was recently changed](https://github.com/quintel/etengine/commit/115c186e6cb34d5473fb85705a03adc804ad3b37) to return something different when a node or object being queried does not exist – as was the case when querying this CHP node. [I've tweaked it again](https://github.com/quintel/etengine/commit/cb56679e4936a901d23f19119e889871a4795681) to not break backwards compatibility with this query.

PR #3847, opened 2021-09-27
Update total cost of electricity production

lottevanvlimmeren:
I have removed or updates the queries in which the old and removed node `agriculture_chp_supercritical_wood_pellets` was present #3842


mabijkerk:
Some label changes for steel production are also included in this commit, is this intentional @lottevanvlimmeren or should these have been added to another branch?

Issue #3833, opened 2021-09-09
AVI (CHP) CCS missing in some charts


MartLubben:
1. Profitability of dispatchable power plants: `plant_profitability`
2. Properties of combined heat and power (CHP) plants: `chp_properties`

Relevant node: `energy_chp_supercritical_ccs_waste_mix`

mabijkerk:
The Waste incinerator with CCS also seems to be missing from the `Total costs of electricity production` chart. In a blank NL scenario, when I add a Waste incinerator, it shows up in the chart, but when I set CCS to 100% the plant does not show up. It clearly cannot be selected in the 'Waste' dropdown:

<img src="https://user-images.githubusercontent.com/67338510/133574197-bbbdb1d2-9c1f-4a22-9a6e-d3045ca7fe57.png" width="500">

It seems as though at least the Waste CHP and Waste CHP with CCS are also missing from this chart. It would be good to add those too and check for any other missing plants.

PR #3853, opened 2021-09-29
Added AVI (CHP) CCS to Total cost of electricity production chart

lottevanvlimmeren:
AVI incinerator CCS, AVI CHP, and AVI CHP CCS are added to the chart 'Total costs of electricity production'


lottevanvlimmeren:
#3833

Issue #3832, opened 2021-09-09
Kerosene mekko does not show demand in sectors other than transport


michieldenhaan:
The kerosene mekko does not yet show demand for kerosene in households/buildings/agriculture/industry.

<img width="655" alt="image" src="https://user-images.githubusercontent.com/32056448/132645857-2e208fd4-a110-46d7-aa11-0fb166c399bc.png">

redekok:
@lottevanvlimmeren would you have time to pick this up?

PR #3857, opened 2021-10-04
Update mekko kerosene

lottevanvlimmeren:
Kerosene used in other sectors than transport are added to the mekko.


lottevanvlimmeren:
Closes #3832

mabijkerk (approved):
Thanks for picking this up!

PR #3871, opened 2021-12-14
Rename NL2019 dataset to NL, NL to NL2015

antw:


marliekeverweij:
@antw I am wondering now why we choose to rename the `nl2019` to `nl`. I think nl2019 is more clear, then you're sure that it's the 2019 dataset. 

Sorry for the (very) late input on this PR!!

antw:
My feeling was that `{country_code}` should represent the latest version of our modeling. Otherwise some – especially those who use the API – are likely to continue using `nl` to create scenarios, unaware that the dataset they are using is old.

marliekeverweij:
Thanks for your reply @antw. Do you say that it is better for our API users to keep the `nl` dataset alive? I would prefer to disable / remove the `nl` dataset and request API users to use `nl2019` instead

antw:
> Do you say that it is better for our API users to keep the nl dataset alive?

I believe so. For two reasons:

1. Including the analysis year is not consistent with the rest of our datasets. We have `uk` not `uk2015`, `fr` not `fr2012`. The change should not be too confusing to API users, since the scenario data *includes* the start year:

    ```js
    {
        "id": 12345,
        "area_code": "nl",
        "start_year": 2015,
        "end_year": 2050
    }
   ```

2. Most importantly, API users may have their workflows broken if the `nl` dataset suddenly disappears.

In the longer term, my preferred approach would be for API users to create a scenario by providing the region name `nl` *and* the start year, and if they don't provide a year we'd default to the latest.

```json
{ 
    "area_code": "nl",
    "start_year": 2019
}
```

marliekeverweij:
Okay, I understand the reasons! Especially the second. 

The first reason doesn't really go for all datasets anymore. Almost all country datasets are now generated through ETLocal and their dataset name is `{geo_id}_{country_name}`. Most ETDataset-datasets have been removed, except for: the Dutch datasets, `dk`,`de`and `eu`.

marliekeverweij:
> In the longer term, my preferred approach would be for API users to create a scenario by providing the region name nl and the start year, and if they don't provide a year we'd default to the latest.

👍 ✨

AlexanderWirtz:
Just weighing in on this discussion:

For transparency and clarity, I believe it would be good to eventually state the year explicitly in the dataset label. We just found out, for example, that for NSWPH Guidehouse was using 2019 datasets for all countries but `de, nl and dk`. Of course, they could have paid a little more attention, but as ever more people create scenarios in bulk through the API, the starting year is less explicitly visible. 

Anyway, since this will have a huge impact on their scenario outcomes and the 2015 datasets are not as good as the 2019 ones, it might be good to come up with a more explicit solution to avoid thing kind of stuff in the future.

kaskranenburgQ:
@mabijkerk Can we close this? 
 I think  a new effort must be made, since there are a lot of conflicts at the moment.

PR #3881, opened 2022-01-04
Innovations flex pvt

lottevanvlimmeren:
Addition of heat pump+pvt and the addition of the roof area chart for households.


mabijkerk (approved):
Updated `pvt-panels `to `PVT panels` in English and `PVT-panelen` in Dutch. Other than it looks good.

PR #3903, opened 2022-02-10
CCS refineries & chemical other sector

marliekeverweij:


marliekeverweij:
It seems that this chart doesn't balance, I guess refineries & chemical have to be added @lottevanvlimmeren do you have time to fix it?
Screenshot 2022-02-10 at 09 15 35

mabijkerk:
For the existing industry sectors where CO2 capture is possible in the model we have researched the capture potential. Are these researched for the newly added CO2 capture sectors as well?

lottevanvlimmeren:
The capture potential has not been research for the refineries and chemical other sector. Instead, the capture potential for the food and paper sector of 85% has been used.

mabijkerk:
> The capture potential has not been research for the refineries and chemical other sector. Instead, the capture potential for the food and paper sector of 85% has been used.

Thanks for the explanation. Given that these values are not researched, what is the reasoning for making the capture sliders available in the main ETM front-end instead of only making them available for external coupling?

marliekeverweij:
Interesting question and I actually do not have the answer. We might want to hide these sliders for ETM users, because we are not sure about the potential. Let's hide them in ETModel and place a note why. Then we save the work and can revert it easily :) @lottevanvlimmeren or @mabijkerk , could you take care of that?

mabijkerk:
I'll hide these new sliders from the ETM front-end.

mabijkerk (approved):
🙌

PR #3921, opened 2022-03-08
Add shortage to Electricity Sankey

mabijkerk:
In https://github.com/quintel/etsource/pull/2644 a number of fixes have been made to the Electricity Sankey. One of these is that shortages due to blackouts have been added to the Sankey queries. This PR adds blackouts to the Sankey.


mabijkerk:
@antw shortages have been added to the Sankey as "blackouts", but perhaps it would better visually when the flow has a gradient like the losses flow has as well. Would it be possible to add this gradient to the blackouts flow and would it be possible to add a "losses" label to the loss flow?

Screenshot 2022-03-08 at 13 12 35

marliekeverweij (approved):

PR #3925, opened 2022-03-15
Demand side response - load shifting in industry

lottevanvlimmeren:
Addition of demand side response in industry: load shifting of electricity is now available for the industrial sectors metal, chemical, and other.


mabijkerk (approved):
Nice work!

Issue #3779, opened 2021-07-28
Update categories in total costs chart


marliekeverweij:
One of the deliverables of the costs improvement project is to update the current cost chart in the ETM. The categories are no longer fully 'mutually exclusive' and for each cost item a subjective choice has been made which category fits best. Examples of choices that have been made: does hydrogen used for heat production fall into the category 'Heat production and networks' or 'Hydrogen'? Do the costs for a power plant with CCS fall into the 'Electricity' or 'CCUS' category? 

![Screenshot 2021-07-28 at 14 35 57](https://user-images.githubusercontent.com/32833996/127323071-9997e99c-fef8-41f1-982e-d2499dc244ad.png)

We'd like to move to the more consistent and transparent chart format that has also been used in the II3050 study:
![Screenshot 2021-07-28 at 14 39 38](https://user-images.githubusercontent.com/32833996/127323485-96ce1946-ea0f-4a4e-b8e8-597e241d85f2.png)

Below you can find a description for each category:
- Gebouwen/installaties (Buildings/installations)
  - Total costs of heating technologies _within a sector_ (gebouwgebonden!) excluding fuel costs
  - Including costs for insulation in houses
- Energiebronnen en import (Energy sources & import): 
  - Costs of primary energy carriers
  - In this category we will show the nett import/export costs, so we will subtract the costs of exported fuels (see [this issue](https://github.com/quintel/etmodel/issues/2204)). 
  - This can be quite tricky for fuels for which the price is different per hour (like electricity).
- Productiemiddelen (Production resources?)
  - Total costs of collective / non-building technologies (producing electricity, hydrogen, heat etc.) excluding fuel costs
  - This category will also include heating technologies for industrial heat network.
- Flexibiliteit en opslag (Flexibility & storage)
  - Total costs of all flexible technologies (P2G, P2P, P2H, dispatchable _gas_ plants). 
  - Distinguish between elektrolyzers: P2G (flexible demand) is part of "Flexibility & storage", dedicated elektrolyzers in combination with wind/solar is part of "Production ..."
- Infrastructuur E/G/W (Infrastructure)
  - Total costs for heat, gas, hydrogen and electricity network
- CO2 netten & CCUS. 
  - Note: this category will be renamed to "CO2 netten, CCUS, compensation", as it will also include the CO2 ETS costs
  - So it will include investment costs for carbon capture, CO2 infrastructure, storage of CO2 and CO2 ETS costs
  
[...]

PR #3927, opened 2022-03-21
Costs update

lottevanvlimmeren:
Updating the cost calculation method: the costs of the energy system are now calculated using newly added cost groups. The cost categories are adjusted; the costs charts are updated accordingly. The fuel costs are now calculated implicitly by subtracting export from production and import of the primary energy carriers.


DorinevanderVlies (approved):

lottevanvlimmeren:
Solves #3779

PR #3999, opened 2022-08-24
Resolve double-counting final demand industry in ETM-CTM coupling

marliekeverweij:
Goes with https://github.com/quintel/etsource/pull/2739


mabijkerk:
Since we actually input natural gas into the energy system, not methane, shouldn't we name it accordingly?

Screenshot 2022-08-25 at 14 20 25

marliekeverweij:
Good idea! I will adjust it

mabijkerk (approved):

PR #4004, opened 2022-09-08
What's new September

mabijkerk:


marliekeverweij:
Looking good!

Few points that I have changed:
- the order of the messages: start with the update of the transition path charts tool
- add the word "nuclear" to the header of the SMR, I think not many people know that these are nuclear plants
- other minor changes

marliekeverweij (approved):

PR #4019, opened 2022-10-11
Making curves for electric ships & planes modifiable

marliekeverweij:
Goes with https://github.com/quintel/etmodel/pull/4019


antw (approved):
:+1:

PR #4027, opened 2022-11-03
Add analysis of flexibility on different timescales

mabijkerk:
This PR enables the analysis of the need for flexibility on a short and long timescale. It does so for hydrogen and network gas by applying a moving average to the residual load curves. The resulting curves are analysed in a new table.

This goes with:

- ETEngine: https://github.com/quintel/etengine/pull/1291
- ETSource: https://github.com/quintel/etsource/pull/2797


marliekeverweij:
Nice work @mabijkerk !

A few comments on the text in the new slide:
- I miss why the table only focusses on gas & hydrogen and not on the other carriers
- I miss an explanation on what a moving average is. Maybe we can link to the documentation if it's stated there?
- I think that a link to the residual load curve chart is interesting, then users really see what's going on: before & after the moving average!

mabijkerk:
> Gas & hydrogen

I would say because of the project scope? I'll mention more explicitly that the analysis is only done for network gas and hydrogen.

> Moving average

I think this is a bit inefficient for us to add documentation for: Wikipedia should be clear enough for any user interested in the concept of moving averages.

> Residual load curve

Good point. I tried to do so by linking to the residual load curve slide, which opens the chart when you follow the link. I also included a toggle in the table which allows you to switch between both charts. For clarity I will add it to the table description.

PR #4040, opened 2022-11-17
Improving labels electricity mix (general & P2G) and tekst ammonia slide

marliekeverweij:
Labels of these charts below have been updated

Screenshot 2022-11-17 at 14 24 41


mabijkerk:
Why not name it "Electricity from import" or "Imported electricity"?

mabijkerk (approved):

PR #4075, opened 2023-02-01
Fix local global pages to work with the changes caused by identity

noracato:
The `compare` functionality (know as local-global) was not working anymore. This mainly happened to changes within the app regarding the new identity stuff.

The index page where you pick your scenarios was relatively easy to fix, but the page where the comparison happens will require more work. Basically, the compare functionality now needs to know about the rest of the app (to use the correct authentication protocol), where before it was more of a standalone thing.

I think it will take me a few more hours to figure out how to make the compare part of the rest of the app - maybe this even means some rewriting of code, causing it to take up half a day or so.

We could merge this index page fix first, before the deploy, and fix the other page after, or I can spend the rest of my day on fixing the comparison page. Let me know what you think @ChaelKruip & @mabijkerk

### DIFF
--- a/app/assets/javascripts/compare/metric.js
+++ b/app/assets/javascripts/compare/metric.js
@@ -1,4 +1,4 @@
-/* globals $ _ Backbone I18n */
+/* globals $ _ Backbone App I18n */
 
 (function() {
   'use strict';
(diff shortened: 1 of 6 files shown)


mabijkerk (approved):

ChaelKruip:
👍

PR #4092, opened 2023-03-24
Discard my Transition Paths

### LINKED ISSUE #4084: Allow users to archive their MYC
After #4083 we can add the option to trash the multi year charts.

We have to think if it would be better to have a seperate trash can. Or to share the trash can with the saved scenarios, but distinguish within the trash can (eg columns, or different color schemes).

- [x] Discard logic in model and controller
- [x] Discard from my scenarios view
- [x] Add to cron job
- [ ] Add discard option to API? Or keep just keep delete like we have now?
- [ ] Do we want to show in the MYC app that this scenario is in the trash?

noracato:
The same discard functionality we have for saved scenarios is now also available for multi year charts.

Two questions remain open:
- Do we want to add a discard option to the API? Or keep just keep the hard delete like we have now?
- Do we want to show in the MYC app that this scenario is in the trash? This could be quite some work.

An impression of the trash can, which is now shared between scenarios and TP's:

Screenshot 2023-03-24 at 14 22 48

---

Closes #4084

### DIFF
--- a/app/assets/stylesheets/_scenarios.sass
+++ b/app/assets/stylesheets/_scenarios.sass
@@ -262,12 +262,16 @@ table.default
         pointer-events: initial
         .button
           background: none
+          &:hover
+            color: #000
+          &.destroy:hover
+            color: darken($error-red, 10%)
       &.discarded .big-icon
         padding: 8px 8px 2px 8px
         color: #666
         border: 1px solid #eee
         border-radius: 24px
-        margin: 10px 10px 10px 0px
+        margin: 10px 10px 10px 0
       a.title, .check input
         // title link and checkbox are clickable for the entire container.
         &::after
@@ -281,6 +285,22 @@ table.default
         span
           padding-left: 0.25rem
           font-size: 0.875rem
+      .sub-row
+        display: flex
+      .type-tag
+        color: $landing-gold
+        font-weight: normal
+        border: 1px solid $landing-gold
+        border-radius: 24px
+        padding: 1px 10px 0 8px
+        display: flex
+        margin-left: 10px
+        margin-top: -2px
+        font-size: 0.75rem
+        line-height: 1.2rem
+        span.icon
+            margin-top: 2px
+            padding-right: 4px
       .check, .scenario-info, .buttons
         padding: 0.5rem
       & > :first-child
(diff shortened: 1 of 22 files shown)


mabijkerk:
Useful new feature @noracato! Below my feedback (somewhat broader scope than just the delete MYC feature, hopefully that's okay).

**General feedback**
Screenshot 2023-03-24 at 17 08 22

- The save icon of the "Save current scenario" could be clearer and align more with the other icons: the other icons are line-icons, while this one is a filler-icon
- I might add a "Save scenario" text to the save icon: the other tabs in "My Scenarios" all have an icon and with texts, so it seems counterintuitive that save scenario does not have text
- I'm not sure which scenario I'm saving here: I'm accessing "My scenarios" from the homepage, so I don't have any scenario open yet. Why do I get the save scenario option?
- I think I would prefer it if each box would have the same size, currently "My Transition Paths" is broader than "Trash"

**Delete MYC feedback**

- I don't see an icon appearing over an MYC scenario when I hover above it with my mouse, like it does for the regular scenarios. I'm therefore not sure how to actually delete an MYC scenario
- For the trash can, it might be useful if the scenario information has a type that says "my scenarios" or "my transition paths". I find it difficult to tell the difference only by the two different icons

noracato:
Thanks for your feedback @mabijkerk!

> For the trash can, it might be useful if the scenario information has a type that says "my scenarios" or "my transition paths". I find it difficult to tell the difference only by the two different icons

I changed the label styles in the trash to be more clear:

Screenshot 2023-03-27 at 12 05 41

> I don't see an icon appearing over an MYC scenario when I hover above it with my mouse, like it does for the regular scenarios. I'm therefore not sure how to actually delete an MYC scenario

I can't reproduce this - with you it also works now, right?

> The save icon of the "Save current scenario" could be clearer and align more with the other icons: the other icons are line-icons, while this one is a filler-icon
I might add a "Save scenario" text to the save icon: the other tabs in "My Scenarios" all have an icon and with texts, so it seems counterintuitive that save scenario does not have text

This is only for admins. Because we have an extra tab, the original button that says 'save current scenario' does not fit. Hence I replaced it with an icon with a tooltip with the original text last week. Normal users are not affected.

> I'm not sure which scenario I'm saving here: I'm accessing "My scenarios" from the homepage, so I don't have any scenario open yet. Why do I get the save scenario option?

Could be a scenario you accessed hours before. I think we keep the session alive for quite a while.

> I think I would prefer it if each box would have the same size, currently "My Transition Paths" is broader than "Trash"

Personally I don't mind it. But this would require a small redesign of the whole page because it simply does not fit 😅

PR #4097, opened 2023-04-04
Filter featured scenarios on homepage

noracato:
To accommodate for the upcoming 16 II3050 scenarios that will be featured we need a filter on the homepage for it no to become a mess.

I quickly made a filter for end years, as it was a simple and easy way to make the lists a bit shorter.
Screenshot 2023-04-04 at 15 30 15

A tab will be automatically created for each new end year that is featured.

I also included a 'last updated at' so our users (and we) can keep track of the up-to-dateness of the featured scenarios. But we can also leave that if you like.

There is now a small introduction to go above the tabs. If one of you could review this text that would be great, you can find it here:
- [English](https://github.com/quintel/etmodel/blob/c2d47ec4d73438545e07a3c0c27601b187e68789/config/locales/en_intro.yml#L16-L18)
- [Dutch](https://github.com/quintel/etmodel/blob/c2d47ec4d73438545e07a3c0c27601b187e68789/config/locales/nl_intro.yml#L16-L18)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -31,3 +31,4 @@
 //= require load_pico_embed
 //= require copy_link_box
 //= require header
+//= require featured_scenario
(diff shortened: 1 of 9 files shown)


noracato:
To be more in line with the new styling I updated this PR.

Screenshot 2023-09-08 at 17 15 42

Charlottevm (approved):

PR #4104, opened 2023-04-14
Improve scrolling experience by fixing menus

### LINKED ISSUE #4100: Improve interactive scaling of frontend
When I open the ETM on (vertically) half of my screen, the frontend is just slightly too wide:

Screenshot 2023-04-11 at 12 12 12

This means that, in order for me to see and use the entire chart, I have to scroll to the right, which causes a white gap to appear in the navbar:

Screenshot 2023-04-11 at 12 12 45

For a short-term solution, it might be useful if the navbar (the top blue bar with features and the dark grey bar with the area and end year) is fixed in place. This would be similar to the dashboard bar, which is displayed properly.

This solution would however still leave part of the selection menu on the left inaccessible. A more structural solution would be to interactively scale the frontend of the model.

### LINKED ISSUE #4101: Adjust position of free space when scrolling up
If I am at the top of the page on LinkedIn, scrolling further up will cause some free space to show **between** the content and the navbar. In the ETM, the same action will cause free space to show **above** the navbar. See the difference in the screen recording below.

It would be cool if we could implement a similar feature as LinkedIn in the ETM. What do you think @noracato @Charlottevm?

https://user-images.githubusercontent.com/67338510/231139298-f0f6a7f3-d93b-48c0-8f7c-9ad1fabbb5be.mov

noracato:
Adds `position: fixed` to all menus.

Closes #4100
Closes #4101

### DIFF
--- a/app/assets/stylesheets/_general_layout.sass
+++ b/app/assets/stylesheets/_general_layout.sass
@@ -23,14 +23,20 @@ body#model
   min-height: calc(100% - 57px)
   min-width: 1000px
   position: relative
+  padding-top: 57px
 
   .flex-wrapper
     align-items: stretch
     display: flex
     flex: 1
+    padding-top: 42px;
 
   #sidebar
     width: 209px
+    position: fixed
+    height: 100%
+    background-color: #fff
+    z-index: 98
 
     & > .sep:last-child:after
       // Adds some space below the last item in the sidebar to ensure that a scrollbar appears as soon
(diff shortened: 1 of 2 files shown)


mabijkerk:
Awesome! One detail is that with the new implementation the items on the navbar are positioned on top of each other when I choose a smaller screen. See on the left the new implementation and on the right the current implementation on pro:

Screenshot 2023-04-14 at 13 05 54

Is there a way to prevent the items on the navbar from stacking like this?

mabijkerk (approved):
🔥

PR #4114, opened 2023-05-15
Transparency of couplings

### LINKED ISSUE #4110: Show `coupled` if scenario was coupled
Add a new item to the scenario nav to show if the scenario was coupled.

image

After implementing quintel/etengine#1340 this `coupled` parameter should be available on the scenario via the API

- [x] Add coupled param to Scenario
- [x] Show status in nav (no dropdown yet)
- [x] Add dropdown with documentation (no uncoupling yet!!)

### LINKED ISSUE #4111: Show if disabled slider was disabled because of coupling
When a scenario was coupled and sliders have been disabled because of it, show an icon with these sliders:

image

After quintel/etengine#1341 this information (possibly including the coupling name) will be availabe thought in `inputs` endpoint of ETE.

- [ ] pick up this extra param and set it on the input
- [ ] show the icon

### LINKED ISSUE #4112: Uncouple a scenario
image

This requires a simple "reset" sent to the affected inputs from ETModel, OR some new "coupled" : "false" param we can set via the ETEngine AP that will solve the uncoupling there internally.

noracato:
This PR adds:
- A new action in the scenario nav showing if a scenario was coupled (closes #4110)
- Shows if a slider was disabled becaue of the coupling (closes #4111)
- Allows for uncoupling of the scenario (closes #4112)

Note that most test are failing because the new API params are not yet on beta. When quintel/etengine#1342 is merged, we can update the tests cassettes.

Impresssions:
Screenshot 2023-05-12 at 10 43 41
Screenshot 2023-05-12 at 16 58 45


noracato:
Thanks for your review! You made some very good points.

> Can we fit it across two lines instead of one?

Of course

> Should we change the text 'Documentation' to 'What does this mean?' and add it to the same text box as 'This scenario has been coupled with an external model'? Moreover, can we somehow clarify that this links to an external website? 

Will do!

> The icons seem to take a very long time to load for the first time, I'm not sure if this can be helped

That's because they are loaded as an image - like the ones in the sidebar. I can check if we can do it with more native `svg` drawing. Will look a bit more messy in the code, but should improve loading time.

> A solution that I like even better is to add information about coupling to the '?' button, in a similar fashion as for technical and financial properties

I like this a lot!

>  I think it would be more in line with the ETM frontend if it is a similar pop-up as for 'Save Scenario'

I'm using the exact pop up you get when resetting a scenario. I can change the style, but as a consequence the reset pop up then also changes.

> It would have space for text that informs users about the effects of uncoupling and a link to the documentation.

I'll update the popup and will put some text in there. Maybe later you can check and adjust it

mabijkerk:
Changes are looking flashy! One final comment: when I hover above the "What does this mean?" in the button, the overlay makes it a bit difficult to read.

image

Can we improve this somehow? Perhaps my idea of marking the link blue didn't come out as well as I expected 😁

noracato:
Everything you mentioned is now done! Could you add some better translations for the confirmation pop up [here](https://github.com/quintel/etmodel/blob/f3f3589290ec1647a28b793fa6fc840afcf59e23/config/locales/en_etm.yml#L313) and [here](https://github.com/quintel/etmodel/blob/f3f3589290ec1647a28b793fa6fc840afcf59e23/config/locales/nl_etm.yml#L316)?

When we're happy, we can first merge quintel/etengine#1342, and I'll rerecord the testing cassettes here.

mabijkerk (approved):

PR #4127, opened 2023-06-27
Add agriculture to p2h costs slider

mabijkerk:
Agriculture p2h boilers were missing from the investment costs of p2h boilers. This PR adds them.

Goes with https://github.com/quintel/etsource/pull/2925


redekok (commented):
Quick remark: wouldn't it make more sense to link to the ETM sections where you can install power-to-heat? Instead of the sections where you can change the willingness-to-pay? I guess that would make more sense to me (but maybe that's personal?).

mabijkerk:
Valid point, I was a bit surprised at first as well. I changed it based on your suggestion.

PR #4140, opened 2023-08-30
Emit postMessage when performing a request in litui mode

thomas-qah:
For a full **What?**, **Why?** and **How?** of the changes in this PR please see the description here:
https://github.com/quintel/multi-year-charts/pull/41

TL;DR:
To implement the third solution as suggested [here](https://github.com/quintel/multi-year-charts/issues/36#issuecomment-1660565860) this PR makes ETModel emit a [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) when it fires a request while in `liteui` mode. Other applications, such as the MYC, can listen for this postMessage event to alter their UI as required.

Closes https://github.com/quintel/multi-year-charts/issues/36

### DIFF
--- a/app/assets/javascripts/lib/app.coffee
+++ b/app/assets/javascripts/lib/app.coffee
@@ -190,8 +190,7 @@ class @AppView extends Backbone.View
   handle_ajax_error: (jqXHR, textStatus, error) ->
     console.log("Something went wrong: " + textStatus)
 
-    if globals.env == 'development' ||
-        (globals.env == 'staging' && globals.debug_js)
+    if globals.env == 'development' || (globals.env == 'staging' && globals.debug_js)
       body = $('body')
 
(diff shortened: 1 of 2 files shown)


noracato (commented):
Works great!
Maybe this is also the moment to fix the styling of the iframe:
```
_liteui.sass l:8-20

  #wrapper, #content
    min-width: initial
    padding-top: 0

  #content #active-content
    padding-bottom: 20px

  #content .flex-wrapper
    padding-top: 0

  #content #sidebar
    height: 100%
```

thomas-qah:
> Maybe this is also the moment to fix the styling of the iframe

@noracato I played around with it a little bit yesterday but couldn't immediately fix it, so was planning to save it for later.

Thanks for the suggested fix! Also works great :) Commited it right away!

PR #4151, opened 2023-10-06
Add brief clarification on scenario compatibility

### LINKED ISSUE #4145: Update text on compatibility when viewing scenario description
The text on the scenario's compatibility can be a bit ambiguous. 

Screenshot 2023-09-28 at 14 50 25

It's unclear if the scenario is being kept up-to-date or not and if these mentioned latest updates have had any influence. We should update this text to be more clear:
- Explain what the current description means: is the scenario still compatible or not?

thomas-qah:
## What?
This PR adds a brief line explaining what it means if the scenario version is different from the current model version.

## Why?
To make the user aware of the possible implications of older scenario versions.

## How?
See changes.

Closes #4145

### DIFF
--- a/app/helpers/scenario_helper.rb
+++ b/app/helpers/scenario_helper.rb
@@ -143,11 +143,14 @@ module ScenarioHelper
   def previous_version_scenario_warning(scenario, current_release_date)
     return if scenario.updated_at >= current_release_date
 
-    t(
-      'scenario.warning.message',
-      last_updated: l(scenario.updated_at.to_date, format: :long),
-      release_date: l(current_release_date.to_date, format: :long)
-    )
+    <<~WARNING.chomp
+      #{t(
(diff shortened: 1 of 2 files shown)


mabijkerk:
Two points of feedback:

1. The Dutch translation seems to be missing
2. I'm not sure this solves #4145, as I see it, this PR only adds a more extensive explanation. What the issue describes is that users do not know if their scenario is being kept up to date or not. I feel the goal of the issue would be to make clear to the user if keep_compatible is `on` or `off`. Do you agree @noracato?

To expand on 2. I think that the we have the following options, that we want to clarify to our users:
- Keep_compatible is set to `false` and there has been a Deploy: the scenario is not changed, but there is a risk that it is broken
- Keep_compatible is set to `true` and there has been a Deploy: the scenario may be changed, but there is no risk that it is borken

PR #4153, opened 2023-10-18
Disallow setting old scenarios as active for a saved_scenario

thomas-qah:
## What?
This PR makes it impossible to set a previously used `Scenario` as the active scenario for a `SavedScenario`.

## Why?
We no longer want to allow our users to manually revert a saved scenario to an older version/scenario. For this specific use-case we will add a new dedicated endpoint to our API, such as `api/v1/saved_scenario/:id/revert_to_scenario/:scenario_id` or so, to do this in a more explicit and controlled manner.

## How?
When updating a `SavedScenario` the given `scenario_id` is checked against the `scenario_id_history`, and an error is returned if it is found.

Closes quintel/etengine#1320

### DIFF
--- a/app/models/saved_scenario.rb
+++ b/app/models/saved_scenario.rb
@@ -127,7 +127,7 @@ class SavedScenario < ApplicationRecord
     incoming_id = params[:scenario_id]
     add_id_to_history(scenario_id) if incoming_id && scenario_id != incoming_id
 
-    # "Manual versioning" by the user, e.g. setting an earlier used scenario-id
+    # "Manual scenario reverting" by the user, e.g. setting an earlier used scenario-id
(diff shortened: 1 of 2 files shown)


app/models/saved_scenario.rb
```diff
@@ -127,6 +127,17 @@ def update_with_api_params(params)
     incoming_id = params[:scenario_id]
     add_id_to_history(scenario_id) if incoming_id && scenario_id != incoming_id
```
noracato, line 128: I think we can move this validation into the `add_id_to_history` method: that would make the validation available for all other code that updates SavedScenarios. Then you also don't need the extra validation step in the service anymore.
Would you mind adding a short test for it as well? Thanks!

Issue #4116, opened 2023-05-23
Non-energetic energy-use in 'other' industry sectors cannot be changed


Charlottevm:
Currently users are unable to influence **non-energetic energy demand** attributed to the **other industry** sector. This means that, for example, if there is a demand for coal in the present users are unable to change or remove this in the future.

To resolve this we could 'easily' duplicate the sliders which influence energetic demand for non-energetic demand.

What are your thoughts @mabijkerk & @ChaelKruip ?

mabijkerk:
I'm not sure what you mean by 'duplicate'. Adding inputs that can be used to adjust the non energetic demand in the other industry seem like a good idea to me.

Charlottevm:
@mabijkerk, I guess there are two options: 

1. we can either keep the current sliders but make sure they influence the energetic _and_ non-energetic demand 
2. or create ('duplicate') similar sliders which influence the nonenergetic demand seperately

![afbeelding](https://github.com/quintel/etmodel/assets/74194356/04f22ad4-d7a4-46bf-9601-7caeb003e027)

mabijkerk:
Adding them to the existing sliders does allow us to keep the front-end a bit more clean. However, changing the sliders will also change all existing scenarios and users may not have had the intention to change the non-energetic demand.

KoenvanB:
To do:

- split size slider into size slider for energetic and non-energetic. Use use:energetic as intersection property
- add input in etsource for non-energetic sliders
- add accordeon + sliders including texts and translations in etmodel

@MartLubben @mabijkerk

mabijkerk:
Overview of work to be done:

- [ ] Add inputs for non energetic shares on ETSource @KoenvanB 
- [ ] Split input of sector size into energetic sector size and non energetic sector size (new on ETSource @KoenvanB 
- [ ] Add inputs to ETModel, add translations and groups @KoenvanB 
- [ ] Add a migration that duplicates the value for sector size of the old input to the new inputs for energetic sector size and non energetic sector size on ETEngine @noracato

PR #4160, opened 2023-11-01
Update sliders in Industry other with non-energetic options

KoenvanB:
Energetic slider for cokes added

Non-energetic sliders added:

- Coal
- Cokes
- Biomass
- Network gas
- Oil
- Hydrogen

Corresponding chart Final energy demand in the other industry updated:
-  Chart also shows non-energetic use of hydrogen and (non-)energetic use of cokes;
- 'coal' labels in chart now correspond to what the coal slider sets, namely: coal and coal gas (previously, the coal slider sets coal demand but the chart showed coal, coal gas _and cokes_ demand)

Solves https://github.com/quintel/etmodel/issues/4116


mabijkerk (approved):
LGTM

PR #4170, opened 2023-11-18
Add multi-user support to scenarios - frontend

### LINKED ISSUE #4159: Multi user scenario support: UI implementation
The effort for this issue involves creating a new form in the front-end, when viewing the details/settings for a specific scenario under My Scenarios. This new form will allow a user to manage the access rights of other users to his or her scenarios.

The concept design for this looks as follows:
Image

The following additional details will be worked out for this issue in regards to the design above:
- A short sentence explaining the concept of scenario 'ownership' will be placed above the user table. A clickable info `🛈`-icon will be placed behind it, which will toggle the visibility of a brief body of text below it which will take the user to the documentation where it's explained what each user role does,
- When clicking the `Invite (new) user` button we will use the existing `Save your scenario` modal as a basis. This will contain a text input where an email address can be entered and a dropdown input where a role can be selected, together with 'Add' and 'Cancel' buttons,
- When clicking the remove `🗑`-icon behind a coupled user we will use the existing  `Uncouple scenario` modal as a basis. This will show a brief 'Are you sure'-text with 'Yes' and 'No' buttons

thomas-qah:
Closed in favor of the PR that combines front- and backend: #4172


thomas-qah:
See description

PR #4172, opened 2023-11-22
Add multi-user support to scenarios (front- and backend)

### LINKED ISSUE #4157: Multi-user scenario support: join table, model adaptation, tests (backend)
The effort for this issue involves adding a new 'roles-rights' system in the backends of `etengine` and `etmodel` that manages access to scenarios by users. The following changes will be made to create such a system:

- Adding a join-table between `(Saved)Scenarios` and `Users` to persist the user's role,
- Adapting the codebase to use this join-table instead of the current `user_id`/`owner_id`

Finishing this will fulfill part of the work to be done for model usability, activity 1.2.1.

### LINKED ISSUE #4159: Multi user scenario support: UI implementation

### LINKED ISSUE #4164: Multi-user scenario support: backend logic for new user management
This issue covers the work that needs to be done to support to new UI that's being built to manage user roles/access to scenarios in issue #4159:
- Handle and tests requests to add, change or remove user roles for scenarios
- Add support for user invitation links and e-mail sending

thomas-qah:
## What?
This PR adds multi-user scenario functionality to ETModel.

## Why?
Some of our clients would like to be able to share scenarios and collaborate on them together with other users. A way to do so is by being able to link multiple users to one scenario and have the ability to give them different roles. This PR does just that.

## How?
Below you can find the most important changes made to make multi-user scenario support possible in ETModel:

### Backend
- Addition of a new model:
  - `SavedScenarioUser`: This model serves as the coupling link between SavedScenarios and Users and symbolizes the role that a user has in relation to a given scenario. For the model to work a `saved_scenario_users` table was also added. This table contains all said information.
- Addition of`'roles` to the `User` model:
```
  ROLES = {
    1 => :scenario_viewer,
    2 => :scenario_collaborator,
    3 => :scenario_owner
  }
```
- Changes to `CanCan abilities` to make sure a viewer can only read, a collaborator can only change and a owner can do everything to a scenario
- Addition of two new controllers:
  - `SavedScenarioUsersController`: This is the 'normal' controller. It handles:
    - Rendering the list of users currently linked to a scenario as a table on the `SavedScenario` show page
    - Rendering the New `SavedScenarioUser` modal shown when a user is added to a `SavedScenario`
    - Rendering the 'confirm destroy' modal shown when a user is about to be decoupled from a `SavedScenario`
    - Creating, updating and destroying `SavedScenarioUsers`
    - When a `SavedScenarioUser` is created/updated/deleted, it performs a call towards ETEngine through new service classes (mentioned below) to make sure the user-role-scenario relations remains the same over there.
  - `API::SavedScenarioUsersController`: This is the controller handling API requests towards `/api/v1/saved_scenarios/:id/users/`. It handles requests for:
    - GET /: Returns all `SavedScenarioUsers` for a given `SavedScenario`
    - POST /: Handles adding multiple `SavedScenarioUsers` in one go. Expects an array in the form of: `[{user_id:, user_email:, role:}, {...}, ...]`
    - PUT /: Handles updating multiple `SavedScenarioUsers` in one go. Expects an array in the form of: `[{id: / user_id: / user_email:, role:}, {...}, ...]`
    - DESTROY /: Handles destroying multiple `SavedScenarioUsers` in one go. Expects an array in the form of: `[{id: / user_id: / user_email:}, {...}, ...]`
- New services to keep `SavedScenarios` (ETM) and `Scenarios` (ETE) in Sync:
  - `CreateAPIScenarioUser`, `UpdateAPIScenarioUser` and `DestroyAPIScenarioUser`
  - These service classes call towards ETE's API endpoint and make sure the user-role-scenario relations that exist in ETModel also exist in ETEngine.
- New logic to make a user's coupling to a scenario definitive:
  This is the mechanic I'm least pleased with. Currently, a `SavedScenarioUser` is added through its email address. ETModel does not have access to user email addressed however, so it cannot verify immediately if a user exists. The current implementation checks at several moments if a user was added/invited to a new scenario:
  - When the user logs in
  - When `ScenariosController::index` is called
  - When `SavedScenariosController::all` is called
  - When `SavedScenariosController::show` is called (e.g. when a SavedScenario is loaded)

### Frontend:
- Various new views:
  - A new `SavedScenarioUsers` management page for a given SavedScenario
  - New modals for creating `SavedScenarioUsers`, and confirming to remove them
- New javascript and style changes to accomodate the new features

Closes #4157.
Closes #4164.
Closes #4159.


noracato:
Hi! I took a first look today, and haven't tested it thoroughly yet - will do that next week with your guide! There were a few small bugs that I had to fix along the way (the discarded scenarios controller wasn't updated, etc.) But otherwise it worked fine!

The UX/UI felt a bit cluttered to me. Would have liked to sit together and go through it, but there's no time sadly. So I quickly updated everything. A few things that would still be great as an addition (but this is probably for another time/project):
- when removing a user, add a cancel button in the popup
- when you are the only user, do not allow role changes or deletion (gray out, no hovers)
- when you invite a new user and leave the email blank, the error message should say: "please fill in an email address"
- when the user is you add `(you)` behind their name
- allow collaborators and viewers to see who are the owners in the main scenario view

Don't think we are still going to pick these up, but this could be a good place to note them down

noracato:
I was checking the part of downgrading your own role more deeply today. I found that in the UI, it's possible to make yourself a collaborator instead of owner and still be able to change the roles of other users. Even after refreshing the page.
This has two reasons:
- The SavedScenario and user permissions are all preloaded and not reloaded on role changes
- The Javascript part has no validations or user feedback

I fixed the js part in my last commit. The Rails part of reloading the permissions is still open, as well as two TODO's I left in the code.

Another question is what we do if a user downgrades their own role. My proposal:
- downgrade to `viewer`: redirect to scenario page
- downgrade to `collaborator`: grey out role changes to others. Redirect only on refresh. On next visit do not allow

The greying out of the roles is already implemented in the last commit.

On last thing we should do, is to always show the names of the owners next to the scenario. As a viewer it would be good to know who was sharing the scenario with you. Let's see if there is time. What do you think @mabijkerk?

**TODO list**
- [ ] Fix Rails side of reloading permissions
- [ ] Check the other two TODO's in the code (search for `TODO: NORA:`)
- [ ] Redirects on self role changes
- [ ] _OPTIONAL_ tag owners in scenario info

thomas-qah:
Hi Nora!

Thanks for checking this, updating the layout and adding the feedback/remarks/to-do's so far! I think it looks much, much better now 🙂

Regarding the issues related to 'downgrading your own role' described in your most recent comment: thank you for catching those loop-holes. I addressed them as follows:
- Added an extra check for `SavedScenario` ownership in the `load_saved_scenario_user` method in the `SavedScenarioUsersController`. This causes the check to happen on `update`, `confirm_destroy` and `destroy` actions,
- Added redirects from `/saved_scenarios/:id/users` to `/saved_scenarios/:id` if a `SavedScenarioUser` is not an Owner.
- Return only owner users in the `owners` array in the output of the `SavedScenario.as_json` method
- Add owners next to 'last updated' under SavedScenario title on show page and users page
I think that should fulfil to-do 1, 3 and 4 in your comment. I left one of your comments from to-do 2 as I'm not sure you'd consider that fixed now. Lets discuss/check that together.

Regarding the to-do's in your first comment, I noticed you fixed some of those points already. I fixed the rest, so I think we should be good now. Let's also discuss/check this together.

noracato:
Hi! Thanks for fixing the issue!

I'll start my code review now. The bullets and TODO's that were marked as optional (all the bullets in my first comment, and the last one in my second comment) were still up for discussion with me and Mathijs. I don't have time to review them, so could you please move your commits containing them to a new branch?

I'll check the comment you left in the code and will remove it when it's indeed fixed!

noracato:
Hey, I just tested it out and the exact same problem still occurs... If I make myself collaborator and refresh, I see myself as collaborator and can still change the roles of the other members.

I'll fix it.

Was also wondering why you always fix the tests in a separate commit in the end, and not run them while developing? And  if you could explain why you decided to add an API controller in your last commit?

PR #4173, opened 2023-11-28
Adding translations and flag of efate

Charlottevm:


redekok:
LGTM!

Two minor remarks:

1. Efate is now categorized under "Provinces - Netherlands"
2. The custom flag is replaced by an EU one when I launch a new scenario (but maybe that's just my local machine doing weird stuff..)

PR #4183, opened 2023-12-15
Introduce more fine-grained load-shifting

thomas-qah:
## What?
This PR adds support for setting up more fine-grained load-shifting by our users.

## Why?
The currently existing load-shifting functionality caused too much of a peak in the demand for the electricity system. The user needed a way to spread out this demand more evenly over time.

## How?
By adding sliders to the UI, including localized descriptions.


mabijkerk (approved):
Minor changes, most notably I changed the input and output capacity labels to load increase and reduction respectively, as this is more intuitive to our users.

PR #4185, opened 2023-12-22
Add local household storage optimization as an option for household battery

noracato:
Adds a cute radio button to select the new option for using forecasting locally: forecast only on the residual load curve for households.
Screenshot 2023-12-22 at 10 02 40

As well as adding the residual load curve for the household sector to the flexibility_hourly_carriers_inflexible_imbalance chart


mabijkerk (approved):
⚡

PR #4187, opened 2024-01-03
 Change of slider text to example

kaskranenburgQ:
Changing slider text of ATR to the example used for the documentation slider text example


mabijkerk (changes requested):
I have made a few minor changes. Always remember to add or update the nl locales, those still need to be changed. Before doing that, let's have a short discussion about the text as well - there are some points I'm still in doubt about.

PR #4195, opened 2024-01-23
Improvements for heat modelling changes

mabijkerk:
This PR updates and implements the households space heating deficits table.

Goes with https://github.com/quintel/etsource/pull/3004


redekok:
Tiny little nitpick: -0 numbers appear in the deficit table. Negative deficits are not possible, are they? Is this minus sign due to rounding then?

image

redekok:
In this scenario (II3050v2 - National Leadership with a few adjustments) it seems as if the absolute deficits from the table do not sum up to the (total) number in the bezier chart. 

Also, I'm not sure how to read the relative deficits. Should all percentages sum to 100%? Or should the percentages per housing type sum to 100%? Or per construction period?

image

mabijkerk:
- Deficits do not sum up: noted, this is something we have to fix
- Deficits are unclear: changed the labels, we should still add description to the table
- Negative deficits: noted, this is something we have to fix

redekok (approved):
LGTM!

PR #4208, opened 2024-02-29
Add propane to network gas mix

KoenvanB:
This PR features several minor front end changes to include propane as a component of the network gas mix:

- Adds an input slider for the share of propane in the network gas mix;
- Adds propane to the network_gas_mix chart and renames the chart label to 'Gas mix in gas network'
- Adds propane to the network_gas_production chart using a flat profile (see corresponding ETSource PR)


mabijkerk (changes requested):
Looking good, I made some minor changes to the slide text and found a chart that is not working properly. It would be nice if we could combine the natural gas, green gas, propane slider set with the natural gas (gaseous), LNG slider set. @KoenvanB could you open an enhancement issue for this?

- [x] The Future energy import chart does not seem to be working as it does not show propane import
- [ ] Open enhancement issue to merge slider set of 3 (natural gas/green gas/propane) and 2 (natural gas/LNG) to a single set of 4 sliders (natural gas/green gas/propane/LNG)

Issue #4211, opened 2024-03-04
Geothermal should be added to the Sankey Energy flow overview


MartLubben:
Link: https://pro.energytransitionmodel.com/scenarios/1097001 
<img width="1397" alt="image" src="https://github.com/quintel/etmodel/assets/33415156/2c2226f7-8c4f-4441-acca-1c879791394b">

kaskranenburgQ:
This issue is a duplicate of issue #[2998](https://github.com/quintel/etsource/issues/2998) of the ETSource repository.

kaskranenburgQ:
As far as I can see the queries for this flow are already created, it just needs to be added in the sankey.

PR #4214, opened 2024-03-07
 Adding geothermal to heat in energy flow overview sankey

kaskranenburgQ:
Closing issue #4211. 

Queries were set, added this to the Sankey.
Tested the results with the graph "District heating supply and demand per temperature level" in a scenario with geothermal must run production in LT, MT & HT heat.

Results can be found underneath:
Screenshot 2024-03-07 at 11 19 36

Screenshot 2024-03-07 at 11 20 27

Both flows in the graphs are equal with the geothermal source node:

Screenshot 2024-03-07 at 11 36 11

@mabijkerk I think I remember this query was not added for a reason but can't remember why at the moment. Do you have an idea?


mabijkerk (approved):
Nice work! I added geothermal for electricity as that appeared to be missing as well.

kaskranenburgQ:
Nice addition! I took the liberty to also fix [this](https://github.com/quintel/etsource/issues/3000) issue concerning the missing Dutch labels in the Sankey.

PR #4220, opened 2024-03-19
Addition of BECCUS to ETM

kaskranenburgQ:
With this update, the option to model BECC(U)S plants with the ETM becomes possible.

For the ETModel repository this means the following changes have been implemented:

- Addition of sliders which set the capacity of biomass plants that only produce power.
- Addition of sliders for dispatchable biomass plants.
- Addition of sliders for CCS for all biomass plants.
- Addition of sliders for Full load hours for must-run biomass plants.
- Addition of these new biomass plants to relevant graphs.
- Addition of slider voor captured biogenic co2 price.
- Creation of new graph called : " Net effect of demand and supply of CO2 on emissions".


config/locales/interface/input_elements/en_costs.yml
```diff
@@ -176,6 +176,17 @@ en:
...
+    costs_captured_biogenic_co2:
+      title: CO<sub>2</sub> removal price
+      short_description:
+      description: |
+        The CO<sub>2</sub> removal price is defined as the price that is received for
+        captured CO<sub>2</sub>. This slider sets the future CO<sub>2</sub> removal price (in €/tonne).
+        The start value is set as 0 €/tonne. <br/><br/>
+        Logically, <strong>the value of the CO<sub>2</sub> removal price should at least be
+        set equal to the CO<sub>2</sub> price</strong> (price for emitting CO<sub>2</sub>). If the
+        CO<sub>2</sub> removal price is set lower than the CO<sub>2</sub> price, it would be
```
mabijkerk, line 188: This description and should be updated to match the **captured biogenic CO2 price**. This also goes for the nl locales.

mabijkerk (changes requested):

mabijkerk:
I think the frontend of the Captured and Imported CO2 slide in CCUS section is quite cluttered. I would prefer a more clear section with multiple slides. These would be the proposed slides:

- Overview (optional)
- Capture of CO2 in energy sector
  - DAC
  - Power plants
  - Hydrogen plants
- Capture of CO2 in industry sector
- Import of CO2

Additionally, the full load hours for must-run CCS biomass can go to the Costs & efficiencies section.

mabijkerk:
It seems to me that a new series needs to be added to both costs charts (detailed and overview). This series should specify the costs of captured biogenic CO2.

Screenshot 2024-03-21 at 11 19 21

mabijkerk (approved):
Nice work y'all!

mabijkerk (changes requested):
For a blank nl2019 scenario, there is demand and supply of CO2, which is clearly shown in the table view of the new waterfall chart:

Screenshot 2024-03-22 at 13 56 03

However, when I look at the chart view, this supply and demand does not appear:
Screenshot 2024-03-22 at 13 56 53

noracato:
> Adding any new flows does work, but still does not make the existing flows visible.

This is entirely expected behaviour.

The waterfall chart by default shows the change between present and future. In your table view you see that present and future are the same, so there is a difference of 0 shown in the chart. When you then change the future capture to export, you can see the changes reflected in the chart.

What was your idea of what the chart has to show? There are also options to show only data of the future, or of the present instead of showing the difference.

mabijkerk (approved):

PR #4236, opened 2024-04-05
Transit flows cost and spatial aspects

kndehaan:
The modelling of costs and spatial impact of transit flows has been improved, covering:

- Costs of required import facilities of energy carriers, added to the cost export and to the detailed annual cost chart
- Spatial requirements of required import facilities of energy carriers, presented in a new table of future spatial requirements

Goes with:
- https://github.com/quintel/etsource/pull/3029
- https://github.com/quintel/documentation/pull/179
- https://github.com/quintel/etmodel/pull/4236
- https://github.com/quintel/etengine/pull/1416
- https://github.com/quintel/etdataset/pull/996


config/locales/interface/output_elements/en_import_export.yml
```diff
@@ -33,3 +33,11 @@ en:
...
       short_description: 
       description:
+    space_requirements_transit_flows:
+      title: Land use of import infrastructure per carrier
+      short_description:
+      description: |
+        This chart presents the imported volume per energy carrier and the 
+        future space requirements of import facilities to process the imported
+        volume. This gives an indication of the spatial impact to facilitate transit
+        flows (import and export) of energy carriers.
```
mabijkerk, line 43: Import is not necessarily only for transit flows: import might also just be for domestic use. More importantly though, I would expect to read here what components included in the land use. Can we clarify this?

kndehaan: I have altered the description:

-  further specified what is included in the required space
- rephrased impact of required space for facilitating import, rather than facilitating transit flows

mabijkerk: 🙌🏻

Issue #4243, opened 2024-04-22
Labels for solar thermals in buildings should be updated


mabijkerk:
For buildings, solar thermal panels cannot supply 100% of hot water demand (as for households), but at most 13% of space heating demand. See the input statement `buildings_space_heater_solar_thermal_share`:

```
# Solar thermal panels can provide a maximum of 13% of year heat demand. Setting slider at 100% means that 13% of heat demand comes from solar thermal and 87% from the main heating technologies in buildings.

- query =
    EACH(
      UPDATE(EDGE(buildings_space_heater_solar_thermal, buildings_useful_demand_for_space_heating_buildings_present), share, (USER_INPUT()*0.13)/100.0),
      UPDATE(EDGE(buildings_space_heater_solar_thermal, buildings_useful_demand_for_space_heating_buildings_future), share, (USER_INPUT()*0.13)/100.0)
      )
```

The frontend of the [Space heating](https://energytransitionmodel.com/scenario/demand/buildings/space-heating) section for buildings however has the following structure:

<img width="400" alt="Screenshot 2024-04-22 at 16 59 50" src="https://github.com/quintel/etmodel/assets/67338510/eeabfc3e-7a49-4216-83a9-604557e0a9b2">

There are two issues here:

- [ ] The group label states "space heating and hot water" for the solar thermal panels, which in this case is applied incorrectly and should be fixed
- [ ] There should a label stating "of potential" for the solar thermal panels in addition to the "% of heat demand" label, to clarify to user that they do not apply to the total heat demand

KoenvanB:
Solved in PR 4245

PR #4245, opened 2024-04-24
Issue4243

KoenvanB:
This PR solves #4243 by

- renaming the accordion for the 'hot water' section;
- attaching the proper units to each section.

Additionally, the 13% hot water demand maximum for solar thermal panels in the Buildings section is now mentioned in the slider text.


kaskranenburgQ (changes requested):
I see the nl_locales are not updated yet, could you add these?

KoenvanB:
Thanks for checking, I have corrected the Dutch translation as well.

kaskranenburgQ (approved):

PR #4251, opened 2024-04-30
Show and alter scenario version history

noracato:
For a long time, we have been keeping the saved scenario history hidden. Now it's time we finally show it to the public!

Users can now
- View the historic versions of their saved scenario's
- Open historic versions
- Restore historic versions
- Add a description to an historic version
- See who last altered a historic version

We also decided to give the whole 'my scenarios' page a small facelift to increase usability. But there could be much more improved there.

Screenshot 2024-04-30 at 13 27 04

---

Still to do:
- [x] Review texts and create Dutch translation 
- [x] Routing (when refreshing history page it should stay on the history page)
- [x] Add API actions

Nice to have:
- [x] Back button / close history button
- [ ] Possible re-adding of (country) flag
- [ ] 'Read more...' for long descriptions
- [ ] When opening an old version, title should be kept with something like _'version 04-05-2024'_ appended

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -328,7 +328,7 @@ GEM
     minitest (5.22.2)
     multi_json (1.15.0)
     multi_xml (0.6.0)
-    mysql2 (0.5.3)
+    mysql2 (0.5.6)
     net-imap (0.4.10)
       date
       net-protocol
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -33,3 +33,4 @@
 //= require copy_link_box
 //= require header
 //= require featured_scenario
+//= require saved_scenario_history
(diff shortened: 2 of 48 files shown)


kndehaan:
Nice work! I checked the first nice to have box since there is already button for closing version history, therefore already implemented right? 

Furthermore, I changed locales and added Dutch translations. After discussion with Mathijs, I also changed the background colour of the drop-down menu back to landing gold (orange). 

Regarding the history versions layout, it looks quite different on my PC than on Nora's screenshot:
image

This is something that should be looked into @noracato 

Other than that, I have no other comments.

kndehaan (changes requested):
Nice work! I checked the first nice to have box since there is already button for closing version history, therefore already implemented right?

Furthermore, I changed locales and added Dutch translations. After discussion with Mathijs, I also changed the background colour of the drop-down menu back to landing gold (orange).

Regarding the history versions layout, it looks quite different on my PC than on Nora's screenshot:
image

This is something that should be looked into @noracato

Other than that, I have no other comments.

mabijkerk (approved):

kndehaan:
Layout is improved for Chrome, only when the description is longer than a couple sentences it becomes a bit messy, is that easy to fix? 
image

noracato:
Yes, I'll give it a quick fix and merge this! Thanks for noticing!

kndehaan (approved):

PR #4261, opened 2024-05-06
Update energy mix visual

kndehaan:
The energy mix visual was outdated and is hereby updated. Updating the visual encompasses the following changes:

- Updating underlying queries of 'Opgave per thema 2050' components
- Updating underlying queries of 'Energiebronnen 2050'
- Locales improvements
- Incorporate ETM logo

Goes with:
- https://github.com/quintel/etmodel/pull/4261
- https://github.com/quintel/etsource/pull/3045

### DIFF
--- a/app/views/energy_mix/_reverse.html.haml
+++ b/app/views/energy_mix/_reverse.html.haml
@@ -226,10 +226,18 @@
     bodem (500 m. diepte) en grote ondergrondse boilervaten.
 
 #disclaimer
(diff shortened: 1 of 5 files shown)


mabijkerk (changes requested):
Good changes for now. Could use a thorough update at some point, for example it contains references to Klimaatmonitor, it is only available in Dutch, it only contains hydrogen import as synthetic gas etc.

Two open questions:

- [x] Determine whether we should hide this for non-NL regions
- [x] Add warning that it will be replaced by 2025

PR #4262, opened 2024-05-07
Setup a red warning banner to tell users about the hydrogen migration

noracato:
Hi! Here's the basic banner. I didn't put the translations in there. Currently is just English in the HTML.

You can update the text here:

### DIFF
--- a/app/views/layouts/etm/_banner.html.haml
+++ b/app/views/layouts/etm/_banner.html.haml
@@ -1,0 +1,24 @@
+.header-banner
+  From
+  %span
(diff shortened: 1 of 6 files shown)


app/views/layouts/etm/_banner.html.haml
```diff
@@ -0,0 +1,24 @@
...
+    discouraged.
+  %a.banner-info
+    Learn more...
+
+.header-banner.header-banner-info
+  .banner-info-box
+    .stick-right
+      %a.close
+        Close
+    It's all about hydrogen
```
noracato, line 16: This is the text in the popup!

PR #4263, opened 2024-05-08
Improving Beccus what's new text

kaskranenburgQ:
This PR improves the text of the What's new page for BECCUS.


mabijkerk:
@kaskranenburgQ I have made a few minor changes in the EN locales. Could you pick up the following:

- [ ] Review my changes and change where necessary
- [ ] Update the NL locales to match the resulting EN locales
- [ ] Add one or two images to the What's new: I would at least add a screenshot of a costs chart with captured biogenic CO2 costs, perhaps a merit order chart with the various biomass (CCS) plants would be useful as well

kaskranenburgQ:
I chose not to add the merit order graph, since I feel that it would not be easy to show the impact of the captured biogenic price to the price of the Beccus plants to the user. I feel it would be a cluttered picture at best, do you agree? @mabijkerk

PR #4264, opened 2024-05-10
Fixing broken links

kaskranenburgQ:
This PR aims to remove the broken links within ETModel.
Most broken links existed because of the two ways slidertexts can be written in the model: with or without the "|" at the start. 
While writing a slider text without the "|" it is necessary to put a "\" before the url. The link will break when a "|" is put before the text. 
Thus, this PR primarily focused on fixing these links.


mabijkerk (approved):
🚀 🚀 🚀

PR #4285, opened 2024-06-07
Save scenarios in a collection and open them in MYC

noracato:
**!!!This PR includes an update from Rails 7.0 to 7.1!!!**

This PR introduces the scenario collection. A rework of the Transition Paths that you could already find in your 'My Scenarios' view.

A Collection can be made through the API by using the existing transition paths API. The only difference is that now you can also add Saved Scenarios, and not just Scenarios, by supplying an array of `saved_scenario_ids`.

Moreover, almost all the functionalities of the API (except `edit`) are now available in the interface.

---

Screenshot 2024-06-07 at 13 48 56
You can still see all your collections and transition paths. And are able to create a new interpolation OR a new collection.

**NOTE** Let's think of a naming convention here! Interpolation or transition path? What is more clear to the user?

 ---

Screenshot 2024-06-07 at 13 49 53

It's easy to create a new collection!

---

Screenshot 2024-06-07 at 13 50 03
The collection can be viewed and shared. If one of the linked scenarios is private, other will not be able to click the boxes to see the scenarios.

When a user clicks the open button, the MYC will be opened with the latest engine-scenarios linked to the saved scenarios (the latest version).

----

Still todo:

- [x] Dutch translations
- [x] Naming conventions interploation/transition paths
- [ ] Documentation on collections
- [ ] Documentation on changes to the API
- [ ] Add a info button somewhere that explains the new functionality (?)


kaskranenburgQ:
While entering some scenario's in the collection, I encountered the following 'error message'. We knew some translations where missing, however, we might also do something about the green check that appears with this error message?
Screenshot 2024-06-10 at 09 06 30

kaskranenburgQ:
This error seems to occur because an empty scenario is added to my scenario collection. If I select these 4 scenario's:
Screenshot 2024-06-10 at 09 43 11

I see the following:
Screenshot 2024-06-10 at 09 42 37

noracato:
Thanks for testing @kaskranenburgQ. The empty one should be filtered out, I noticed it already before and it should be handled in the code. Do you have a larger picture of the error? I think something else is wrong.

> we might also do something about the green check that appears with this error message

Good catch, I'll change that up!

kaskranenburgQ (commented):
LGTM!

PR #4290, opened 2024-06-17
Hybrid offshore wind

kndehaan:
With this PR the modelling of hybrid offshore wind is added to the ETM. This comes with, among others, the following features:

- Hybrid offshore wind turbine that can generate electricity for HV network or directly for connected offshore electrolyser
- Price-based offshore electrolyser that can receive electricity from hybrid wind turbine or from onshore HV network
- Front-end sliders to set installed (relative) capacities of the components and other relevant specs (e.g. costs, full load hours)a
- 3 new front-end outputs: a bar chart, Sankey diagram and table that provide data on capacities and energy flows of hybrid offshore wind

Goes with:
- https://github.com/quintel/documentation/pull/195
- https://github.com/quintel/etsource/pull/3067
- https://github.com/quintel/etlocal/pull/524
- https://github.com/quintel/etmodel/pull/4290
- https://github.com/quintel/etdataset/pull/1001
- https://github.com/quintel/merit/pull/154
- https://github.com/quintel/atlas/pull/167
- https://github.com/quintel/etengine/pull/1440


mabijkerk:
Electricity consumption by the hybrid electrolyser seem to missing from the chart `merit_order_hourly_flexibility`, `hourly_flexibility_electricity`, `use_of_excess_electricity` and `installed_flexible_capacities`. It also appears to be missing from the table `storage_options`

kndehaan:
I've added the hybrid offshore electrolyser to the `merit_order_hourly_flexibility`, `hourly_flexibility_electricity`, `use_of_excess_electricity`,`installed_flexible_capacities` and `table storage_options`.

Regarding:

- `flexibility_integral_costs`: I think the hybrid offshore electrolyser should indeed be added as an additional category next to regular P2G, will do that.
- `co2_intensity_of_hydrogen_production`: I think P2G should also be added here, but since this chart is already rather 'busy' I would suggest to include the offshore electrolyser in the already present P2G category.

kndehaan:
`co2_intensity_of_hydrogen_production` and `flexibility_integral_costs` are updated now. Only the `flexibility_integral_costs` has some layout issues considering the labels. We could consider abbreviating the labels?

kndehaan:
I'm not sure if it's manageable to get the layout of `flexibility_intergral_costs` right considering labels with P2G offshore added as a new category. 
image

I think if I abbreviate the labels more than this, it becomes rather unclear which categories are presented in the chart (of course we can explain it in the chart description, but the chart on it self should be self explanatory to a certain extend). So I would say either add the hybrid offshore electrolyser to the existing P2G category or omit it from the chart. 

@mabijkerk do you have any suggestions or preferences?

mabijkerk:
> I think if I abbreviate the labels more than this, it becomes rather unclear which categories are presented in the chart

I agree. I see two valid routes: option 1 is to adjust the chart settings so that text overflow is put in a new line and option 2 to split storage into a second chart. Option 1 would require a dev effort and enhances the overall usability of the chart. Option 2 is interesting because it would allow us to split electricity storage into separate technologies.

However, both options require more effort than seems feasible for this project at this point. I would therefore leave out P2G offshore from this chart, change the existing series to P2G onshore for clarity and note the above as a future improvement. What do you think @kndehaan?

kndehaan:
I agree that option 1 and 2 are currently out of this project's scope (I'll note this as future improvement) and therefore agree to leave out P2G offshore and updating existing series' label. Will process this.

mabijkerk:
I see that an image has been added for the What's new page, but no What's new has been written yet. It would be good to update the What's new page (and the What's new bracket, see below). Also note that we typically include a Dutch and English image.

Screenshot 2024-06-18 at 15 53 09

mabijkerk:
It may be good to consistently label "onshore" p2g where relevant. In the chart `sankey_electrical_interconnection` for example, we have the label p2g and offshore p2g. In the chart `source_of_electricity_in_p2g` we do not state that this refers to onshore p2g.

app/assets/javascripts/d3/sankey.coffee
```diff
@@ -615,6 +617,29 @@ D3.sankey =
...
+          {left: 'hybrid_offshore_wind',    right: 'offshore_electrolyser',     gquery: 'hybrid_offshore_wind_to_offshore_electrolyser_in_sankey',      color: '#1f77b4'},
+          {left: 'hybrid_offshore_wind',    right: 'from_offshore_network',     gquery: 'hybrid_offshore_wind_to_from_offshore_network_in_sankey',      color: '#1f77b4'},
+          {left: 'hybrid_offshore_wind',    right: 'curtailment',               gquery: 'hybrid_offshore_wind_to_curtailment_in_sankey',                color: '#1f77b4'},
+          {left: 'offshore_electrolyser',   right: 'onshore_hydrogen_network',  gquery: 'offshore_electrolyser_to_hydrogen_network_in_sankey',          color: '#87cfeb'}, 
+          {left: 'offshore_electrolyser',   right: 'conversion_losses',         gquery: 'offshore_electrolyser_to_conversion_losses_in_sankey',         color: '#cc0000'},
+          {left: 'from_offshore_network',   right: 'onshore_hv_network',        gquery: 'from_offshore_network_to_onshore_hv_network_in_sankey',        color: '#1f77b4'}
```
mabijkerk, line 641: These labels and query names could be somewhat clearer.

mabijkerk:
I would keep the What's somewhat shorter. The main goal of the What's new is to inform users what new functionality has been added and where to find it. By opening the model they can then explore these new features themselves. For explanations of the functionality itself, the user should be referred to the documentation.

Also note that we typically include a Dutch and English version of any image with text.

Don't forget to also update the What's new box on the homepage:
Screenshot 2024-06-20 at 10 49 53

PR #4295, opened 2024-06-26
Separate units column

### LINKED ISSUE #3862: Show table units in a separate column
In the current table display of charts, for example for the chart `CO2 emissions`, the units are displayed in each cell in the chart:

Screenshot 2021-10-26 at 10 53 56

The issue with this visualisation is that 1) it has a lot of redundancy, decreasing the clarity of the table and 2) it is less easy to copy table data to Excel. My proposal would be to change this visualisation, for example by adding a separate column where the units are displayed.

@marliekeverweij what do you think of this idea?
@noracato would this idea be possible?

louispt1:
As requested in #3862, units have now been separated into another column. This is what it looks like:

image

A more robust solution might externalise the js in this _table_template to a helper method. I had a bit of trouble with this because I would need to make the html template a .erb to use the helper methods and there were some rendering issues going down that route.

This works so maybe it's easier to leave it as is 😄 .

Closes #3862


mabijkerk (approved):
I have not reviewed the code itself, that's up to @noracato, but in terms of functionality it works perfectly! Quick work 🏎️

mabijkerk:
@louispt1 what about translations for the word "Unit"? In Dutch this would be "Eenheid".

louispt1:
> @louispt1 what about translations for the word "Unit"? In Dutch this would be "Eenheid".

Fixed

kndehaan:
I also had a quick look, looks very good! Would it be possible to move the "Unit" column to the front? So before the first data column (in this case before "2019" column).

louispt1:
> I also had a quick look, looks very good! Would it be possible to move the "Unit" column to the front? So before the first data column (in this case before "2019" column).

Fixed now :)

kndehaan:
Looking good!

noracato (approved):
Great work!

PR #4303, opened 2024-07-18
simple csv download functionality

louispt1:
This is not fully ready for merging. I'm making a pull request because I'm a little stuck on how to hide/show the download csv button depending on the chart conditions. I know there is a 'can_be_shown_as_table' variable in charts, but it's always undefined by the time the table is being rendered... quite sure I'm missing something obvious here 🙃.

I could implement a solution based on the type of chart being rendered, but it feels like the long way around. Maybe you know how this works @noracato?

Everything else is working quite well - this solution is only for charts that can be viewed as tables currently - you can download the other chart data but it's never really pretty.

P.s. - This method (from **base_chart_view.coffee**) looks like it's for updating the icons depending on conditions? Played around with this but I couldn't even get it to trigger with console.logs, nor find the _download_csv_ element.

```
update_header: =>
    # Skip re-rendering if this chart is about to be replaced.
    return if @model.get('will_replace')

    id = @model.get 'chart_id'
    @$el.data('chart_id', id)
    @$el.find('h3').html(@model.get("name"))
    @$el.attr('data-block_ui_on_refresh', @block_ui_on_refresh())
    @$el.find('a.chart_info').toggle(@model.get('has_description'))
    @$el.find(".actions a.chart_info").attr "href", "/descriptions/charts/#{id}"
    @$el.find(".actions a.zoom_chart").attr "href", "/output_elements/#{id}/zoom"

    if @canDownloadImage()
      # saveAsPNG is exposed by app/javascripts/packs/app.ts
      #
      # Unbind any previously assigned saveAs event to prevent it being
      # triggered multiple times.
      @$el.find(".actions a.chart_to_image")
        .show()
        .off('click')
        .on('click', (event) -> BaseChartView.saveAsPNG(event, App.scenario.get('id')))
    else
      @$el.find(".actions a.chart_to_image").hide()

    @$el.find('.actions a').removeClass('loading')

    @format_wrapper = if @$el.parents(".fancybox-inner").length > 0
      @$el.parents(".fancybox-inner")
    else
      @$el

    @format_wrapper.find("a.chart_format, a.table_format").hide()
    if @model.can_be_shown_as_table()
      if @model.get 'as_table'
        @format_wrapper.find("a.chart_format").show()
      else
        @format_wrapper.find("a.table_format").show()

    @$el.find(".chart_not_finished").toggle @model.get("under_construction")
    @$el.find("a.default_chart").toggle @model.wants_default_button()
    @$el.find('a.show_related').toggle @model.wants_related_button()
    @$el.find('a.show_previous').toggle @model.wants_previous_button()

    @update_lock_icon()
    @$el.find('.actions').show()

    this

```

### DIFF
--- a/app/assets/javascripts/lib/models/chart.coffee
+++ b/app/assets/javascripts/lib/models/chart.coffee
@@ -168,6 +168,18 @@ class @Chart extends Backbone.Model
     _.indexOf(['d3', 'import_export_cwe', 'import_export_capacity', 'import_export_flows', 'import_export_renewables', 'sankey', 'storage', 'target_bar'], @get 'type') == -1
 
   # -- series and values -----------------------------------------------------
+  get_table_data: =>
+    headers = ['Label', 'Unit', 'Present', 'Future']
+    unit = @get('unit') || ''
+    rows = @series.map (s) ->
+      [
+        s.get('label'),
+        unit,
+        s.safe_present_value?() || '',
+        s.safe_future_value?() || ''
+      ]
+    [headers].concat(rows)
+
 
   # raw array of the associated gqueries. Delegates to the collection object
   #
--- a/app/assets/javascripts/lib/models/chart_list.coffee
+++ b/app/assets/javascripts/lib/models/chart_list.coffee
@@ -297,6 +297,12 @@ class @ChartList extends Backbone.Collection
   setup_callbacks: ->
     load_chart = @load
 
+    # Event binding for downloading CSV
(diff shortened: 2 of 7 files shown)


louispt1:
Because 'can_be_shown_as_a_table' is set to true in this class, and then overrided by other derived classes, the condition is always true at this point. So it always shows either the table or chart icon ... except in practice sometimes it doesn't show either. 

image

I'm also a tad confused about whether both the app/javascript and app/assets/javascripts are in play - certainly the assets js is used, but is the other js still used in the etmodel?

PR #4304, opened 2024-07-19
changed margins

### LINKED ISSUE #4244: "ppartementen" should be appartementen in the chart
image

louispt1:
Simple bandaid solution. 

Given more time, I think there'd be a way to dynamically update the margins based on the length of the longest label which would certainly be more robust. Happy to spend more time on that solution if you think it's worth it @noracato, although based on my preliminary investigations it's not as straightforward as it sounds.

closes #4244


noracato (approved):
Less is more!

PR #4316, opened 2024-08-09
adjusted container size for sankey

### LINKED ISSUE #4255: Sankey should fit inside it's outside corners
The sankey should fit inside it's corners. I am not sure what the reason is. 

image

https://beta.energytransitionmodel.com/saved_scenarios/5265

louispt1:
Resized the SVG container slightly - this solves the Sankey issue in @MartLubben's example, hopefully it solves all instances of this problem.
I think the problem at its core is related to the d3 sankey @available_size() method. This PR means the container will always increase to 109% of it's initially calculated height.

Closes #4255


louispt1:
@mabijkerk is this fine to merge? Increases the height of the container by 9% which should handle the cases where the Sankey doesn't fit.

`@svg = @create_svg_container @width, @height * 1.09, @margins`

mabijkerk (approved):
Works, so go ahead and merge! Needs to be rebased though.

PR #4317, opened 2024-08-21
Switching from kaminary to Pagy in ETModel

### LINKED ISSUE #4307: Navigation in My Scenarios is broken
In the navigation pane of My Scenarios, I can switch to the second page. However, once I'm there, the navigation buttons do not give any response. I'm working in Google Chrome.

Screenshot 2024-07-24 at 15 31 01

louispt1:
This PR is to switch fro Kaminari to Pagy in ETModel only. Kaminari still seems to be working quite well in ETEngine so there may not be any reason to change just yet - except for Pagy boasting r.e. their memory usage, speed etc ....
There is a [migration guide](https://ddnexus.github.io/pagy/docs/migration-guide/) which proved fairly useful.

Everything works! The only thing is that the styling is a bit different from how it was before.

Before:
image

After:
image

Edit - made it much closer to the original now.
image

In theory you could customise this forever, so if there is something that would be nice to have or something you really don't like about the current implementation I'm very happy to fiddle with it.

Closes #4307


louispt1:
Having looked at this a bit more, I think it may be affecting the api requests? saved_scenarios_spec and transition_paths_spec are both unhappy - do you know what could be causing this @noracato?

mabijkerk:
This is picked up in  Stable versions.

PR #4322, opened 2024-08-23
CTM-ETM coupling: generic transformation nodes

mabijkerk:
This PR adds generic transformation nodes for external coupling by the CTM:

- `energy_chemical_fertilizers_transformation_external_coupling_node`
- `energy_chemical_other_transformation_external_coupling_node`
- `energy_chemical_refineries_transformation_external_coupling_node`

These nodes have a set of input carriers and a set of output carries that allows a robust range of carrier transformations to be loaded from the CTM through external coupling inputs. Not defined input and output can be defined on the nodes as well, to indicate any carriers that are not in scope of the coupling.

**Note:** the robust range of carrier transformation does lead to potential circularities in the graph. This work package therefore includes changes that allow both hourly and annual methods to compute circularities. For the recursive factors we have opted for a workaround that nettifies disallowed inputs to deal with circularity. This is clearly not a full solution, so 'solving' circularities in recursive factors is still a desirable future improvement.

Goes with:
- Refinery: https://github.com/quintel/refinery/pull/62
- Atlas: https://github.com/quintel/atlas/pull/170
- ETEngine: https://github.com/quintel/etengine/pull/1455
- ETSource: https://github.com/quintel/etsource/pull/3107


kndehaan (approved):
Ready to merge

PR #4332, opened 2024-09-09
Soft uncoupling and coupled curve uploads

noracato:
This PR contains a new interface for soft uncoupling models, based on their coupling group as defined in the new config file.

It also allows for switching coupled user uploaded curves, like must run interconnectors.


config/initializers/coupling_group_wrapper.rb
noracato, line 1: I understand your choice, but I'm wondering if we should have created a new model for it. Like that we would be able to define some helper methods on the model for the views. What do you think @louispt1?

Anyways, there is a big chance this exact feature is short lived. So let's keep it like it is!

louispt1: I agree - a model would make more sense long term! I also thought this might be a 'stop-gap' solution while there are only really two coupled models

louispt1:
LGTM! Goes with [this ETEngine PR](https://github.com/quintel/etengine/pull/1459)

kaskranenburgQ:
In a scenario in which the interconnector is set to must-run, the price of the interconnector is still shown in the graph. 
While effectively the price of the interconnector is zero.
Screenshot 2024-09-10 at 10 51 54
@mabijkerk Do we need to build in an if statement in the queries for the interconnector prices for this graph?

PR #4333, opened 2024-09-11
Clarify terminology for coupled external models

mabijkerk:
Proposal to clarify terminology of coupled model.

- Activate coupled model
- Inactive coupled model ("soft uncoupling", the link is still there, but it's not active)
- Uncoupled model ("permanent uncoupling", the link is broken)

@kaskranenburgQ @noracato please see if you agree with the proposed terminology and check my work. If we push these changes we also need to update the documentation about external models accordingly.

### DIFF
--- a/config/locales/en_etm.yml
+++ b/config/locales/en_etm.yml
@@ -59,12 +59,12 @@ en:
     current_scenario: "Current scenario"
     coupled: Coupled
     coupled_docs: What does this mean?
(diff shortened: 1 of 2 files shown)


kaskranenburgQ (approved):
LGTM!

mabijkerk:
Goes with https://github.com/quintel/documentation/pull/205

noracato (approved):

PR #4351, opened 2024-10-16
CTM-ETM additional carriers

kaskranenburgQ:
This PR adds bio-oil as input-carriers to the industry transformation nodes and adds bio-oil, bio-kerosene and biodiesel as output-carriers to the industry transformation nodes. 

Goes together with: 
[ETsource PR](https://github.com/quintel/etsource/pull/3131)
More details can be found there.


app/assets/javascripts/d3/sankey.coffee
kndehaan, line 1: In the biomass sankey, the label of non-biogenic carriers on the left side is not entirely correct since wood pellet input will be accounted for in this flow. Option is to filter this flow out of the non-biogenic flow and add this to dry biomass flow (similar to the what I've done here for bio-oil https://github.com/quintel/etsource/pull/3131/commits/4f556632e27bb74f9e1cc761c43a45ba6ddc8f5f, however this was to avoid double counting). It's a minor detail, we could also leave it as it is now.

kndehaan (approved):

app/assets/javascripts/d3/sankey.coffee
kaskranenburgQ: I think it's good to review this graph in a meeting, due to it's complexity.

PR #4363, opened 2024-11-01
Adding on shore power supply to slider texts

kaskranenburgQ:
This PR Adds on shore power supply to the slider texts for electric shipping in demand.


kndehaan (approved):
LGTM

Issue #4330, opened 2024-09-02
The chart "net flow of refinery products" seems to be incomplete


DorinevanderVlies:
The left half of the net flow of refinery products seems to be missing

## Live

<img width="614" alt="image" src="https://github.com/user-attachments/assets/bead07cc-5bbc-4df1-800c-9de9b0d326d5">

## Beta

<img width="614" alt="image" src="https://github.com/user-attachments/assets/3ce9d6f6-810c-409e-a412-8f4fdb8deed5">

kaskranenburgQ:
Following a discussion within the modeling team, we are not completely satisfied with the current design: 
As of now, the "non-refinery products" from the "other" and "fertilizers" external coupling transformation nodes are both used as input and output.  This is a somewhat replication of the information in the "Energy flow overview" and not necessary information for this graph.
![Net flow of refinery products 1193372](https://github.com/user-attachments/assets/9ae0bfdb-48c2-462e-a85b-2c64d258a4d5)

Therefore we propose the following changes
- Seperation of other and fertilizers transformation nodes vs refinery transformation node. 
- Oil product outflow of other and fertilizer transformation node on the second column, no indication of inflow of 
- Name-change of "non-refinery products" to "non oil products"
Will pick these changes up myself.

Issue #3131, opened 2019-09-24
Update list of preset scenario's


DorinevanderVlies:
The current list of preset scenario's is outdated.

I propose to remove the scenario's with a red number:
![image](https://user-images.githubusercontent.com/19907658/65509751-7857da80-ded3-11e9-9daf-c0d9927e6564.png)

Curious about your ideas @marliekeverweij @MartLubben @michieldenhaan @redekok @ChaelKruip @jorisberkhout @AlexanderWirtz

AlexanderWirtz:
I am even wondering if we should continue #5. Perhaps you and Joris have already discussed this with Marjan Minnesma? 
In my opinion that scenario needs a fundamental rethink.

DorinevanderVlies:
> I am even wondering if we should continue #5. Perhaps you and Joris have already discussed this with Marjan Minnesma? In my opinion that scenario needs a fundamental rethink.

Good point. We raised this issue last week. Fact is that this scenario is still being used by Urgenda. Therefor I propose to keep it in the presets and at the same time try to get going with a fundamental rethink.

MartLubben:
We still use the example of the Energieakkoord (2023) in lectures, just to show how the costs of carriers affects the total costs of the energy system. I know it is outdated, however can't we make a library where users (including ourselves) can find these scenario's. I think it is still valuable to explore older scenario's. .

DorinevanderVlies:
Closing this issue after a short discussion with @MartLubben .
The old preset scenario's are still available (and can be used for presentations), but they are no longer displayed on the front page

PR #4369, opened 2024-11-21
Redesigning refinery sankey

kaskranenburgQ:
This PR updates the Refinery Sankey by taking in the following changes as described in https://github.com/quintel/etmodel/issues/4330:

- Seperation of other and fertilizers transformation nodes vs refinery transformation node.
- Oil product outflow of other and fertilizer transformation node on the second column, no indication of inflow of
- Name-change of "non-refinery products" to "non oil products"

With the addition of bio oil carriers to the transformation node in #3131 these carriers have also been added to the non oil products for now, since these are (as far as I'm aware) not classified as refinery products. 

Goes together with:
https://github.com/quintel/etsource/pull/3144


mabijkerk (approved):
I recommend we squash these commits.

PR #4370, opened 2024-11-26
Update locales heat demand buildings

kndehaan:
Applied correct English locales, in line with the Dutch locales.


kaskranenburgQ (approved):
LGTM!

Issue #4345, opened 2024-10-01
Clarify coupling in UX for Interconnectors


louispt1:
Separate to #4341, because the interconnector coupling simply de-activates some sliders and replaces the upload module with the new coupling curves, we decided it's not clear enough that a coupling is in effect.

So this issue is to update the curves upload (when coupled) so that it also has the little coupling icon.

![Screenshot 2024-10-01 at 11 50 50](https://github.com/user-attachments/assets/83d9ed2f-cddb-47c6-89a3-dfbfe5d38640)

The icon would go on this part:
![image](https://github.com/user-attachments/assets/025bfc0e-1930-4a2d-b039-bc5dd2f3ce0a)

kaskranenburgQ:
Some other issues concerning the external coupling of interconnectors we found are:

- Uploading a must-run curve returns the error message "An error occured". A refresh of the page indicates that the curve is actually loaded in properly, so the error message is false.
![Screenshot 2024-12-03 at 09 17 03](https://github.com/user-attachments/assets/3555eadf-7a8c-4ad2-94df-bc148d30b809)
- When the external coupling input `external_model_interconnectors` is set to 1.0, all the slide texts for the curves will be set to "Coupled custom curves". Even when the interconnector is not coupled, as seen here:
![Screenshot 2024-12-02 at 15 14 28](https://github.com/user-attachments/assets/cc7693d4-0083-4599-9efe-ff5fbf0faf5e)

- In talks with the client they mentioned that it's NTH to have the possibility to upload the price-curve of the interconnector as well.

PR #4372, opened 2024-12-01
Changes for external coupling of must-run interconnectors

mabijkerk:
See https://github.com/quintel/etsource/pull/3155. I made the following changes:

- Added FLH external coupling inputs to front-end

Notifying @kaskranenburgQ and @louispt1. There are a couple of to-do's still left before we can merge:

- [ ] Depending on the discussion about the capacity profile in https://github.com/quintel/etsource/pull/3155, we may need to update locales
- [ ] When uploading the must-run curve through the front-end, it says that an error occurred, after refreshing however, the curves are actually uploaded
- [ ] After adding the FLH external coupling inputs, the disabled availability and costs inputs are no longer hidden

### DIFF
--- a/config/interface/input_elements/external_coupling_interconnectors.yml
+++ b/config/interface/input_elements/external_coupling_interconnectors.yml
@@ -1,0 +1,204 @@
+---
+- key: external_coupling_interconnector_1_flh_imported_electricity
+  step_value: 1.0
+  unit: 'flh'
+  position: 100
+  slide_key: flexibility_electricity_import_export_interconnectivity
+  share_group:
+  interface_group: full_load_hours
+  coupling_icon: true
+- key: external_coupling_interconnector_1_flh_exported_electricity
+  step_value: 1.0
+  unit: 'flh'
+  position: 105
+  slide_key: flexibility_electricity_import_export_interconnectivity
+  share_group:
+  interface_group: full_load_hours
+  coupling_icon: true
+
+- key: external_coupling_interconnector_2_flh_imported_electricity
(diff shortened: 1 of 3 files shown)


mabijkerk:
Given the limited time available from @louispt1, we're not going to address the second and third issue now. We should add them to https://github.com/quintel/etmodel/issues/4345.

kaskranenburgQ:
At the moment, the capacity slider is also hidden when the external coupling is set to active. This is because the capacity sliders are in the same interface_group (called 'capacity') as the interconnectivity sliders. 
There are at least 2 ways to solve this:

- Splitting the interface groups
- Adding an external coupling slider that sets the capacity (@mabijkerk we discussed this option already, could solve this as well)

kaskranenburgQ:
> After adding the FLH external coupling inputs, the disabled availability and costs inputs are no longer hidden

After reloading the page a couple of times, this can occur and vanish again as well. Seems to be a bug in the front-end.

mabijkerk:
Good feedback @kaskranenburgQ. About the interface groups: let's await what the client prefers and see if we can redesign the frontend afterwards. Any remaining bugs we can add to the existing issue and dev should pick them up when they have time.

mabijkerk:
As stated in https://github.com/quintel/etsource/pull/3155#issuecomment-2511921536, we can close this PR because the FLH inputs are removed. We can then open a new PR to tackle the issue with the interface group and we still need to add the front-end issues to https://github.com/quintel/etmodel/issues/4345.

PR #4382, opened 2025-01-03
Mekko main carriers electricity

kaskranenburgQ:
This PR adds the mekko for supply and demand of the electricity carrier. 

Goes together with: 
https://github.com/quintel/etsource/pull/3169


mabijkerk:
I'm not too sure about the colour of the Gas CHP (see below). Typically we use orange for nuclear in the model, and grey for gas. This seems somewhat inconsistent.

image

PR #4388, opened 2025-01-09
Non energetic emissions split

kaskranenburgQ:
This PR splits the non-energetic and delayed emissions in the co2 chart. 
In addition more documentation about the non-energetic co2 emissions in the chart are added.

Goes together with:
https://github.com/quintel/documentation/pull/213
https://github.com/quintel/etsource/pull/3174


kndehaan (approved):
I did some minor rewriting in chart description, no comments otherwise!

mabijkerk (changes requested):
This PR still fails on Semaphore.

kndehaan:
> This PR still fails on Semaphore.
I ran the test again, now it passes. Don't know why it failed with the last commit...

config/locales/interface/output_elements/en_overview.yml
```diff
@@ -12,15 +12,18 @@ en:
...
-        target="_blank">documentation</a>.
+        This chart compares CO<sub>2</sub> emissions in 1990, the present, and the future.
+        The year 1990 is often used as a reference year in CO<sub>2</sub> targets.
+        <br/><br/>
+        For the CO<sub>2</sub> emissions of 1990, the ETM considers only domestic, energy-related CO<sub>2</sub>
+        emissions and the non-energetic CO<sub>2</sub> emissions from the fertilizer sector.
+        Non-energetic emissions in other sectors and other greenhouse gases (such as methane) are not included.<br/><br/>
+        The start and future year take the same CO<sub>2</sub> emission categories as the 1990 emissions into account. Besides non-energetic
+        CO<sub>2</sub> emissions in the fertilizer sector, the non-energetic CO<sub>2</sub> emissions of final demand of hydrogen and ammonia
+        in the other industry sector are now accounted for (which are assumed to be zero in the 1990 CO<sub>2</sub> emissions).<br/><br/>
```
mabijkerk, line 23: I would change this on two aspects. First, the **primary CO2 emissions** of non-energetic final demand of hydrogen and ammonia are taken into account (so the emitted CO2 during the production). The use itself cannot have CO2 emissions. Second, I would generalise this to state that the primary CO2 emissions of any non-energetic use of hydrogen and ammonia is taken into account - it does not only relate to the industry other sector, there is also non-energetic final demand for hydrogen in the chemical other sector.

kaskranenburgQ: I have revised the text to: 

> Besides non-energetic CO<sub>2</sub> emissions in the fertilizer sector, the non-energetic primary CO<sub>2</sub> emissions  of final demand of hydrogen and ammonia are accounted for, which are assumed to be zero in 1990. The primary CO<sub>2</sub> emissions for these carriers are the emissions that are released during the production of these carriers.

I'm in doubt whether to add an extra sentence that adds that these carriers do not have carbon atoms. 
What do you think @mabijkerk ?

kndehaan: @kaskranenburgQ  I discussed with Mathijs, adding this additional sentence is not necassery.

Issue #4375, opened 2024-12-04
Hybrid oil heatpump missing from graph "Number of residences per space heating technology


kaskranenburgQ:
If I set this in a Northern-Ireland scenario I see that the technology is missing from the chart.
![Screenshot 2024-12-04 at 15 49 44](https://github.com/user-attachments/assets/f4b09905-e156-4389-bb03-0460cb2b5482)

There are queries already inplace for this technology, they can be found on etsource/gqueries/output_elements/output_series/vertical_stacked_bar_number_of_houses_per_space_heating_technology/space_heater_hybrid_crude_oil_heatpump_air_water_electricity/

mabijkerk:
Well-spotted! Would be good to pick this up at some point.

PR #4389, opened 2025-01-09
 Adding hybrid crude oil to number of residences per space heating graph

kaskranenburgQ:
This PR closes https://github.com/quintel/etmodel/issues/4375

It uses labeling and coloring similar to the other labels and coloring of heat pumps and crude oil used in the "Number of residences per space heating technology" graph.

Goes together with:
https://github.com/quintel/etsource/pull/3175


kndehaan (approved):
LGTM!

PR #4391, opened 2025-01-13
Removing external coupling inputs labels for P2H WTP

### LINKED ISSUE #4384: External coupling inputs for WTP do not show up in scenarios
In the II3050v2 scenarios, the external coupling groups for all industry sub sectors are active:

Screenshot 2025-01-06 at 09 15 06

This means that the regular WTP inputs for industry P2H, like `wtp_of_industry_chemicals_fertilizers_flexibility_p2h_electricity`, are deactivated:
```
- disabled_by_couplings = [external_model_industry, industry_chemical_fertilizers]
```

However, the external coupling inputs that should replace them do not show up in the front-end:
Screenshot 2025-01-06 at 09 17 28

This is because they have not been set in the scenario. This however leads to the strange situation where neither the 'regular' nor the external coupling inputs are set. The WTP inputs are therefore completely lacking from the front-end and can no longer be adjusted by the user (in the front-end). @kaskranenburgQ @kndehaan @noracato, how can we address this issue?

kaskranenburgQ:
This PR removes the the labels for the external coupling inputs for power-to-heat willingness-to-pay in the industry.

It removes the individual labels for each of the external inputs as well as the labels for the group in each file.
closes #4384 

Goes together with:
https://github.com/quintel/etsource/pull/3185
https://github.com/quintel/etengine/pull/1488


mabijkerk (approved):
🙌

PR #4394, opened 2025-01-22
Cleaning up ETModel

louispt1:
I tried to clean up as much as possible, including resolving a few code smells and formatting issues.

I suspect, but am not 100% confident, that these files could also be removed:
- Saved scenario helper - focussed on _displaying_ the saved scenario page
- Multi year chart model - Should this also be a dry struct? Or removed?
- cron.rake - Should we put this in MyEtm maybe?

I couldn't resolve the pending spec for the saved scenario controller - I find it a little difficult to know how to test that.

I would also especially like for you to check [this commit](https://github.com/quintel/etmodel/commit/bbd2b620f12c22ef263d23f698d4d84d8bdbc967) as I'm not fully convinced this is what you meant in your TODO. Should the method be on the Current singleton or the Setting model?


louispt1:
@noracato I think this is pretty far along now - only some VCR tests having trouble still (I'm not convinced I understand how to run them properly).

The scenarios controller spec is still on pending - I think that's the last proper one that might need handling.

Still not sure about [this commit](https://github.com/quintel/etmodel/commit/bbd2b620f12c22ef263d23f698d4d84d8bdbc967).

The other thing I'm not sure about is that at the end of app.coffee there is a user.scenarios method - I think that might be necessary for some front end functionality? If that's true, we will probably need an index method on the saved scenarios controller to fetch all the saved scenarios for the current user

noracato (approved):
Thanks a lot!
I fixed some stuff, for instance raising errors in the controllers. We don't want our users to see the site crash when something goes wrong. It's better to send a message to the user, and if we feel like it's important, we can 'manually' send the error to Sentry.

I started to remove the MYC as well, but ran into a lot of assets errors. You can find that branch at `remove-myc`.

noracato:
Btw, I believe the things you mentioned in your comments are already resolved, correct? The commit looks fine!

PR #4401, opened 2025-02-05
Reverted navbar to keep old content in Model rather than redirecting

louispt1:
I reverted the navbar back to the previous implementation - are there any sections (other than the game which is already gone) which need pruning?

A lot of these items redirect to the same page in the original set up.
image

The 'My Scenarios' and 'Transition Paths' sections feel a bit redundant to me given the MyETM 'hub' implementation. However removing them would leave this section pretty sparse. Note: I haven't changed the links yet because I think it's worth discussing whether we should keep these.
image

image

### DIFF
--- a/app/assets/stylesheets/_static.sass
+++ b/app/assets/stylesheets/_static.sass
@@ -1,6 +1,5 @@
 .static-page
   color: #1f2937
-  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol"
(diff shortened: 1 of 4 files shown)


louispt1:
> Thanks for this @louispt1. Some feedback:
> 
> * I think we could leave out the icons in the navbar for Features and Collaborate.
> * We can move API from About to Features
> * We could Dataset Manager from About to Features?
> * I agree that keeping in My Scenarios does not make too much sense, we can leave it out
> * I would leave in the Collections module under Features
> 
> In the future we can look into a more thorough redesign of the navbar, but that's not in scope now.

Nice ideas! I have applied the changes and the PR is now ready for review 👍

PR #4402, opened 2025-02-06
Get rid of flags and current language in locales switcher

louispt1:
More unified with MyETM


louispt1:
@mabijkerk what do you think? I personally don't mind leaving it as it was for the header and removing the flags in the dropdown.

I guess we're splitting hairs a little here 🙂

mabijkerk:
I'm in favour of leaving it out, but it's definitely not a must-have for this project.

PR #4403, opened 2025-02-10
Removing line that sets scenarios back to blank in revert unsaved changes button.

### LINKED ISSUE #4399: Revert unsaved changes button resets the entire scenario instead
The 'Revert unsaved changes' button implies that any unsaved changes that are have been mode to the scenario will be removed and that the scenario go back to the state at the time of the latest save. 

Currently, this is not the case. While pressing the button, the scenario will reset all the changes that have been made to the scenario. 

This effectively means that the scenario will return to a blank scenario.

**Update**: 
With https://github.com/quintel/etmodel/pull/4403, we partly solved this issue. 
reverting unsaved changes will only work when:

1. You save your (non-blank) scenario.
2. Open a different or new scenario
3. Open your saved scenario.
4. Make changes
5. Press the 'revert unsaved changes' button.

If you don't open a different scenario in between, the revert unsaved changes button will revert to a blank scenario since your sessionID will not have changed, causing etmodel to think there is no scenario to revert to.

To fix this @mabijkerk suggests the following: 

>  perhaps we should force a new SessionID anytime you save a version?

kaskranenburgQ:
This PR aims to close #4399. 

The issue was that the "removed unsaved changes" button removed all scenario-settings in one scenario.
This meant that when you would alter a scenario, and wanted to revert your changes a blank scenario would return. 

This PR removes the line of code in etmodel that resets the scenario when this button is pressed, as proposed by @noracato [in this comment](https://github.com/quintel/etmodel/issues/4399#issuecomment-2633043567).

It's important to note that with this fix, reverting unsaved changes will only work when:

1. You save your (non-blank) scenario.
2. Open a different or new scenario
3. Open your saved scenario.
4. Make changes
5. Press the 'revert unsaved changes' button. 

If you don't open a different scenario in between, the revert unsaved changes button will revert to a blank scenario since your sessionID will not have changed, causing etmodel to think there is no scenario to revert to.

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -97,7 +97,6 @@ class Setting
   #
   def reset_scenario
     self.api_session_id = nil
-    self.preset_scenario_id = nil # to go back to a blank slate scenario
 
     %i[network_parts_affected locked_charts].each do |key|
       reset_attribute(key)


mabijkerk (approved):

mabijkerk:
This works as described by @kaskranenburgQ.

> If you don't open a different scenario in between, the revert unsaved changes button will revert to a blank scenario since your sessionID will not have changed, causing etmodel to think there is no scenario to revert to.

This comment shows that it's not yet perfect though. I would suggest opening an issue to further improve the revert functionality - perhaps we should force a new SessionID anytime you save a version?

noracato:
I think you did not understand correctly. It behaves exactly like we discussed. New IDs are already forced for each version since a long time. So I'm closing the related issue again :)

PR #4408, opened 2025-02-17
Use resource server from the identity gem

noracato:
Bumps the Identity gem to a version that supports setting a valid resource server URI (in this case ETEngine) and makes it possible to call a client to the resource server from the gem.

This PR contains updated configs, and the removal of redudant logic.

Still open:
- Support audience claim as array

Goes with:
quintel/etmodel#4408
quintel/my-etm#72
quintel/identity_rails#6


lib/etmodel/tokens.rb
```diff
@@ -0,0 +1,52 @@
...
+    # Fetches and caches the JWK set from the IdP.
+    def jwk_set
+      jwk_cache.fetch('jwk_hash') do
+        client = Faraday.new(Identity.discovery_config.jwks_uri) do |conn|
+          conn.request(:json)
+          conn.response(:json)
+          conn.response(:raise_error)
+        end
+
+        JSON::JWK::Set.new(*client.get.body['keys'])
```
louispt1, line 38: In the Engine you have it like this:
        `JSON::JWK.new(client.get.body['keys'].first.symbolize_keys)`
        
I'm not sure I understand why they're different? Maybe if it's not necessary we can make them the same

louispt1 (approved):
Looks good to me! Only one small question I wasn't sure about

lib/etmodel/tokens.rb
noracato: You're right, in Engine we just grab the only key that is provided, because I couldn't get the symbolising right, the keys were not recognised, and there was only one key anyway. I can have a second look if I can fix it now!

PR #4411, opened 2025-02-19
Align CO2 emissions in GHG chart with CO2 emissions chart

kndehaan:
This PR aligns the energetic and non-energetic CO2 emissions depicted in the GHG emissions chart with the emissions depicted in the CO2 emissions chart. This is achieved by applying the same underlying, robust queries for both charts. 

With this PR on ETModel, queries in the GHG emissions chart used for delayed and indirect emissions are changed to make use of robust queries. 

Goes with https://github.com/quintel/etsource/pull/3216


kaskranenburgQ (approved):
LGTM!

PR #4414, opened 2025-02-20
What's new February

kndehaan:
What's new text and changelog file for deploy February.


config/locales/en_intro.yml
```diff
@@ -19,7 +19,7 @@ en:
     whats_new:
       title: What's new?
       description: |
-        New supply and demand charts added for district heating (HT, MT, LT).
+        New module "MyETM" available. New supply and demand charts added for the central network of electricity and network gas.
```
mabijkerk, line 22: We don't actually call it "MyETM" in the front-end. Does it then make sense to refer to MyETM? Shouldn't we say the scenarios and account page has been updated (or something similar)? This also applies to the rest of the What's new.

mabijkerk:
All in all, I like the new style! General question: isn't it a bit too full? I was wondering if we should perhaps remove the sub bullets for 1990 emissions. Or is it full because we didn't do a deploy in two months and should we just accept it the way it is?

kndehaan:
It's quite full because we didn't deploy in almost two months, so I think for now it's ok to have it this full.

config/locales/en_intro.yml
kndehaan: I will rephrase the intro and What's new!

mabijkerk (approved):

PR #4428, opened 2025-03-03
Save scenario as button works even on first open of a scenario

### LINKED ISSUE #4471: Fix "save scenario as" button on first open of scenario
At the moment the button doesn't work

louispt1:
Reorganised the .haml so the 'save scenario as' button works even for the first save of a scenario.

Closes #4471


noracato:
@louispt1, could you describe the issue this is solving?

louispt1:
This issue has already been resolved

PR #4431, opened 2025-03-06
What's new deploy March

kndehaan:
What's new for deploy March.


config/locales/en_intro.yml
```diff
@@ -19,7 +19,7 @@ en:
     whats_new:
       title: What's new?
       description: |
-        Update of the "My Scenarios" and "My Account" pages. New supply and demand charts added for the central network of electricity and network gas.
+        Introduction of stable versions of the ETM for API users.
```
mabijkerk, line 22: Following my previous comment, I would not mention the stable versions here. Instead, maybe we should just leave something about "My Scenarios" here, since the last Deploy was so recent?

mabijkerk (approved):

PR #4433, opened 2025-03-07
Functional inputs serialisation to csv

louispt1:
Functional inputs serialisation to csv, fetching from engine via passthru to serve in the data export section.

Dutch translation courtesy of ChatGPT.

Using locales for the inputs would be a nice addition, however at this stage it might be too much effort for the marginal gain.

Goes with this PR in etengine: https://github.com/quintel/etengine/pull/1506

Closes [#4448](https://github.com/quintel/etmodel/issues/4448)

### DIFF
--- a/config/interface/slides/data_export.yml
+++ b/config/interface/slides/data_export.yml
@@ -3,6 +3,10 @@
   position: 100
   sidebar_item_key: data_export
   output_element_key: final_energy_demand_per_sector_joule
+- key: data_export_inputs
+  position: 101
+  sidebar_item_key: data_export
+  output_element_key: data_export_inputs
 - key: data_export_energy_flows
   position: 105
   sidebar_item_key: data_export
--- a/config/interface/slides/data_export.yml
+++ b/config/interface/slides/data_export.yml
@@ -6,7 +6,7 @@
 - key: data_export_inputs
   position: 101
   sidebar_item_key: data_export
-  output_element_key: data_export_inputs
+  output_element_key: final_energy_demand_per_sector_joule
 - key: data_export_energy_flows
   position: 105
   sidebar_item_key: data_export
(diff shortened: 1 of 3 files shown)


mabijkerk (approved):
Nice work @louispt1! I made some minor changes. Always remember to add an existing output element (a chart or table) to the `output_element_key` row in the slides config.

noracato (approved):
Looks good! Just one question: why did we make a new thing and not adjust the old CSV download that we had for saved scenarios which included all locale translations?

noracato:
I talked with Mathijs about this, and the main idea was indeed to restore the old saved scenario functionality. Sorry for the extra work you did in the engine.

I kept the old csv [here](https://github.com/quintel/etmodel/blob/master/app/views/saved_scenarios/show.csv.erb). The route and view should be moved from the saved scenarios, to the scenarios controller, referring to something like `@scenario` in the view instead of `@saved_scenario`.  No need for anything in the engine right now.

louispt1:
> I talked with Mathijs about this, and the main idea was indeed to restore the old saved scenario functionality. Sorry for the extra work you did in the engine.
> 
> I kept the old csv [here](https://github.com/quintel/etmodel/blob/master/app/views/saved_scenarios/show.csv.erb). The route and view should be moved from the saved scenarios, to the scenarios controller, referring to something like `@scenario` in the view instead of `@saved_scenario`. No need for anything in the engine right now.

Hey @noracato, sorry I'm a bit confused because I can't tell if you made this comment or the edit on your other comment first - would you prefer I implement the old show.csv.erb method without the min and max or proceed with the new approach with locales?

louispt1:
@noracato I tried to set up the old functionality on this branch: `restore-old-csv-functionality`. Annoyingly the subscripts and quotation marks cause quite some problems with this approach. I tried all sorts of things including stuff like:
```
CGI.unescapeHTML(CGI.unescapeHTML(strip_tags ...)
```
But the csv still doesn't come out perfectly, and not all inputs are included (for example a lot of the emissions inputs are not shown even once set).

Issue #4434, opened 2025-03-07
Revise description of CO2 capture in industry slide


kndehaan:
The slide description of "Capture of CO2 in industry sector" should be revised:
- It should become clear that the sliders set the share of the capture potential of the industry sub sectors. This is mentioned in the slider descriptions, but is also helpful to mention in the slide description
- Fix spelling errors in the Dutch text

PR #4435, opened 2025-03-07
Update CO2 capture industry slide text

kndehaan:
Updates slide text of CO2 capture in industry, solves https://github.com/quintel/etmodel/issues/4434


kaskranenburgQ (approved):
LGTM!

PR #4443, opened 2025-03-20
Adding other sector to Industry and energy section in co2 sheet

kaskranenburgQ:
This PR solves the quick fix as proposed in [ETSource #3217](https://github.com/quintel/etsource/issues/3217) by adding the emissions of the Other sector to the Industry & Energy sectors. 

@noracato I also updated the file: app/views/pages/dataset.html.haml, since the old querie was mentioned in this file, could you tell me whether this is necessary or whether the file is outdated?

Goes together with:
https://github.com/quintel/etsource/pull/3236


noracato (commented):
Hey! Yes, that's correct. The `dataset.html.haml` is the base HTML file for the co2 sheet. Not sure why it's called like that though, slightly confusing 🤷‍♀️

kndehaan (approved):
LGTM!

Issue #4425, opened 2025-02-27
Split energetic and non-energetic flows in other sector final demand chart


mabijkerk:
In most of the final demand charts, like Chemical industry sector, the series are split into energetic and non-energetic (feedstock) demand:

![Image](https://github.com/user-attachments/assets/a73ddac7-90fe-4055-b2b0-e08834003a29)

In the Other sector however, this distinction is not made:

![Image](https://github.com/user-attachments/assets/6e633ede-40c8-4ac5-a62a-20f6b9d0b369)

I would propose that, instead of showing the sum of energetic and non-energetic demand, the Other sector final demand chart should have separate series for each.

kaskranenburgQ:
I created an inventory of all the output series that do not have a subsequent folder on etsource. 

Since there are a lot of them you can view them below this drop down:

<details>
<summary>Output series</summary>
<br>
- buildings_heat_demand_and_production (outdated)
- dynamic_demand_curve 
- electricity_hv_network_load
- electricity_lv_network_load
- electricity_mv_network_load
- electricity_network_load
- emissions_agriculture
- emissions_built_environment (outdated?)
- emissions_indirect
- emissions_industry_energy
- emissions_overview
- emissions_transport
- final_demand_mekko_energetic_mece
- final_demand_mekko_mece
- final_energy_demand_per_sector_joule
- flexibility_hourly_all_carriers_cumulative_imbalance
- flexibility_hourly_carriers_inflexible_imbalance
- flexibility_hourly_electricity_supply_demand
- flexibility_hourly_heat_network_supply_demand
- flexibility_hourly_hydrogen_supply_demand
- flexibility_hourly_network_gas_supply_demand
- flexibility_hourly_p2p_storage
- flexibility_monthly_carriers_inflexible_imbalance
- flexibility_monthly_carriers_inflexible_supply_demand_max
- flexibility_monthly_carriers_inflexible_supply_demand_sum
- flexibility_of_backup_scatter_plot (oudated)
- heat_infrastructure_annualised_costs_ht
- heat_infrastructure_annualised_costs_lt
- heat_infrastructure_annualised_costs_mt
- heat_network_demand_ht
- heat_network_demand_lt
- heat_network_demand_mt
- heat_network_production_ht
- heat_network_production_lt
- heat_network_production_mt
- heat_network_storage_ht
- heat_network_storage_lt
- heat_network_storage_mt
- household_heat_demand_and_production (outdated?)
- hv_demand_curves
[...]

PR #4445, opened 2025-03-20
Splitting energetic and non energetic demand in final demand other sector chart

kaskranenburgQ:
This PR solves #4425 by splitting non-energetic and energetic final demand in the Other sector chart and updates the labels.
In addition, I have removed the queries for carriers that do not have a final demand node in the other sector. 

Goes together with:
https://github.com/quintel/etsource/pull/3237


kndehaan (approved):
I restored the original order of the carrier series in the chart. The general idea behind this order is that the fossil carriers or carriers expected to the decrease in the future are at the bottom of the chart and renewable carriers or carriers expected to increase in the future are shown at the top. 

Other than that, ready to merge!

PR #4446, opened 2025-03-21
Remove unnecessary data

louispt1:
As discussed, a migration to remove data tables which are no longer necessary in the Model. There was some fallout in the spec after the migration, which I cleaned up with minimal overall changes.


louispt1:
> It's not just about dropping the tables, but also removing all remaining views etc. I'd like to split this PR up into dedicated parts: remove everything around collections, everything about saved scenarios, etc.

Ok, yes I may have misinterpreted this back in March - perhaps we should discuss in the sprint how best to tackle this one

PR #4452, opened 2025-03-31
Featured groups years

noracato:
Quick fix!

Screenshot 2025-03-31 at 18 08 58


mabijkerk:
I like the gradient! I do think that this isuse would require further discussion before merging: I'm a bit worried about the amount of space that the featured scenarios would take up when this PR is merged.

mabijkerk:
The issue is still relevant and it is a project within the User interfaces theme. I  didn't know it had a PR. Should we close it but link it for reference when we start the project?

Issue #2025, opened 2016-02-08
Electric vehicles have zero capacity in storage chart in the present (with MO and excess electricity)


RobTerwel:
To reproduce:

- MO on
- inland wind turbines >10000 to make sure we have excess electricity
- flexibility options ordering: EV storage first
- Storage in EVs slide: increase availability to some number

![screen shot 2016-02-08 at 11 04 31](https://cloud.githubusercontent.com/assets/14106387/12882799/4d394bf6-ce54-11e5-839f-b679cfb3f74b.png)

The capacity and annual demand are zero, even though there is an energy flow. Changing the EV share in transport and then resetting it to its initial value does give us a capacity and annual demand:

![screen shot 2016-02-08 at 11 12 59](https://cloud.githubusercontent.com/assets/14106387/12882906/f8889c46-ce54-11e5-93c5-93e75f2a97de.png)

I think this is because there is no number_of_units calculation in the present yet (see https://github.com/quintel/etsource/issues/1038). An ugly (temporary!) way to solve is, is to put the EV `number_of_units` hard-coded and use the value for the Netherlands.

Assigning myself.

PR #4453, opened 2025-04-02
Add version tags to featured scenarios

mabijkerk:
The landing page shows all featured saved scenarios that are listed in MyETM. This includes scenarios that are listed across different versions. Opening a specific scenario will open that specific version, which might not be clear to users. Therefore, version tags are added.

### DIFF
--- a/app/assets/stylesheets/_root.sass
+++ b/app/assets/stylesheets/_root.sass
@@ -227,7 +227,6 @@ body#root
 
   #new-scenario, #preset-scenario
     border: none
-    margin-bottom: 80px
     .header-container
       display: flex
       margin: -15px -20px 0px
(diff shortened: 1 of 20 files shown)


mabijkerk:
The functionality seems to work for #latest and looks good. Two comments. First, I'm not too sure about underscoring the scenario titles. Second, I can't test multiple versions because I can't get etmodel to run due to the CSRF error. Additionally, I'm not sure #2025-01 is running locally.

Screenshot 2025-04-02 at 11 01 06

louispt1:
> The functionality seems to work for #latest and looks good. Two comments. First, I'm not too sure about underscoring the scenario titles. Second, I can't test multiple versions because I can't get etmodel to run due to the CSRF error. Additionally, I'm not sure #2025-01 is running locally.
> 
> Screenshot 2025-04-02 at 11 01 06

Would you prefer bold for the titles @mabijkerk?

louispt1 (approved):
I removed some unused code and changed the underline to bold but otherwise LGTM

PR #4458, opened 2025-04-03
Update labeling flh's wind in the front-end

kaskranenburgQ:
This PR updates the FLH sliders for wind so that it's more clear that these sliders alter the FLH's for production technologies based on wind energy.


kndehaan (approved):
LGTM!

PR #4460, opened 2025-04-04
Make featured scenarios an active resource to help caching

noracato:
By being an ActiveResource, we need less logic for featured scenarios, which increases caching options.


louispt1 (approved):
LGTM I just re-recorded the cassette for the spec that was failing

Closes #4463

PR #4466, opened 2025-04-07
Add subheader 'electricity input' to storage capacity sliders

kndehaan:
For clarification, the subheader 'electricity input' is added to storage technology capacity sliders, to indicate that sliders set the electricity input capacity.


mabijkerk:
LGTM!

PR #4472, opened 2025-04-23
Update What's new April

kndehaan:
Bug fix for aligning storage technologies (https://github.com/quintel/etsource/pull/3247) is added.


mabijkerk (changes requested):
I think we can do without the first sentence, since the last two sentences basically cover everything users need to know.

mabijkerk (approved):

PR #4494, opened 2025-05-15
Update custom curves documentation

kndehaan:
Related to this [PR](https://github.com/quintel/etengine/pull/1565) that ensures that uploaded custom curves with `capacity_profile` are also downloaded again in the same format, there is a need to update descriptions accordingly. 

This PR therefore updates descriptions to:
- Improve clarity about in which format custom curves should be uploaded
- Indicate that the downloaded custom curves are obtained in the same format
- Underline that these curves represent scenario input and are thus not the outcome of scenario calculations

Along the way, some outdated documentation regarding custom curves is deleted or updated where relevant. 

Goes with:
- https://github.com/quintel/etengine/pull/1565
- https://github.com/quintel/documentation/pull/230
- https://github.com/quintel/etmodel/pull/4494


louispt1 (approved):
LGTM but I only checked the English version

PR #4495, opened 2025-05-16
Updated SavedScenarios call to pass 'all' param

louispt1:
Added call scenarios endpoint with the all param for modify profiles section

Goes with: https://github.com/quintel/my-etm/pull/132


noracato (approved):

louispt1:
> I think the download of the csvs in the results page is not yet solved by this. That is a communication issue to etengine regarding tokens I assume.

Which csvs are not working? I can download csvs in the results page when not logged in / in an incognito browser

PR #4500, opened 2025-05-20
Sorting implemented on WTP for flex

louispt1:
Sorting by WTP in the VariableEconomicPerformanceTableView. Had to name the file 'variable economic performance...' because the assets are loaded in alphabetical order...


app/assets/javascripts/lib/views/power_plant_economic_performance_table_view.coffee
kndehaan, line 1: Why is this file deleted? I think this is still needed for the sorting of the economic profitability table of the power plants, which follows the sorting rules of the merit order table.

louispt1: I thought that the profitability of dispatchable power plants table was `plant_profitability` using the PlantProfitabilityTableView - is that incorrect?

Or - what is the table that is no longer being sorted? This line:
`when 'power_plant_economic_performance' then MeritOrderTableView`
should use the sorting from the MeritOrderTableView.

Lmk if I'm missing something or we need to have a quick call!

louispt1:
> Sorting seems to work well and showing the dashes in WTP column if the value is zero as well! Only tested it for the storage table.
> 
> From the code, I think all zeroes in the table should be replaced by dashes, but that is not working yet for the CAPEX. Any ideas why this is not working? To my opinion, it's rather a nice to have for the columns other than the WTP column, so I wouldn't dive too deep into it if it can't be solved easily.

Hey @kndehaan - I took a quick look at this, but I'm not sure how to make zeros in the relevant columns. Could you check if it's working after the latest commit or lmk which inputs I should set / which scenario would be good to work off for this? Maybe I'm using the wrong ETSource branch because it's been a minute and there are a couple 😄

PR #4501, opened 2025-05-27
What's new June

kndehaan:
What's new and changelog added for deploy June.


app/assets/images/pages/whats_new/changelog/202506_changelog_inputs.csv
mabijkerk, line 1: Two questions about the changelog. First is that typically I think we exclude external coupling inputs from the changelog, do you think we should include them? Second is that all the changes are being deployed with the June deploy, so why does it sometimes say that it's part of the February deploy?

mabijkerk:
@kndehaan by using a different format for the internal links, they work both for beta and pro. For example `[Flexibility](scenario/flexibility/flexibility_net_load/hybrid-offshore-wind-components)`.

app/assets/images/pages/whats_new/changelog/202506_changelog_inputs.csv
kndehaan: Regarding your first question about external coupling inputs, I think it depends on what has been changed in the inputs whether to include or exclude them in the changelog. Even though only a few people use these inputs, it might be helpful to them to know that changes are done. For example, I remember a client experienced issues with `external_coupling_industry_metal_aluminium_technical_lifetime` which caused the scenario to break due to a set (and allowed) value of of 0. It might be convenient for the client to know that this bug is solved (by changing the minimum value to 1). 

I do agree however that we can be more strict in whether to include or exclude certain changes related to external coupling inputs. 

Regarding your second question, I remember that some time ago that we discussed to include the changed inputs of the previous deploys of a calendar year in the CSV. So that in December's deploy, you have the complete list of changed inputs of the entire calendar year. We haven't had any changelogs this year except for Ferbuary, that's why now only inputs of deploy in June and February are mentioned in this changelog.

mabijkerk: I still have a slight inclination to leave out the external coupling inputs: they can clutter up the document and should only be used by expert users who know what they're doing. If you disagree, fine to leave it as is.

> I remember that some time ago that we discussed to include the changed inputs of the previous deploys of a calendar year in the CSV

Right, makes sense.

kndehaan:
> @kndehaan by using a different format for the internal links, they work both for beta and pro. For example [Flexibility](scenario/flexibility/flexibility_net_load/hybrid-offshore-wind-components).

Good observation! I used the `<a></a>` format to enable that the link opens in a new window (`target="_blank"`) instead of leaving the what's new page when clicking on the link. But I now changed the links so that it opens in the same environment (e.g. `<a href="/scenario/demand/households/appliances" target="_blank">`)

app/assets/images/pages/whats_new/changelog/202506_changelog_inputs.csv
kndehaan: Yeah on second thought, I agree that the external coupling inputs are a quite advanced expert feature. I've hereby omitted them from the changelog.

PR #4511, opened 2025-06-23
Upgrade ruby to 3.3.5

### LINKED ISSUE #4468: Update ruby version to at least 3.2
As ruby 3.1 is losing support, we should upgrade to at least version 3.2.

noracato:
I tried to fix the other branch again, but with no luck. So I thought to just start again.

This one has no removal of old stuff, no updates to any css, packs, packers and other bothersome figures. Just a plain version upgrade. Seems to work for some reason...

Only updated vcr and pry.

Closes #4468


noracato:
Note: I did not yet check if the image is building

PR #4515, opened 2025-06-30
What's new July 2025

kaskranenburgQ:
This PR adds the what's new for July 2025.


app/assets/images/pages/whats_new/economic_viability_en.png
mabijkerk, line 1: @kaskranenburgQ I like that the table is rather short. If possible, it might be nice though to replace it with a table where (some) power plants make a profit? The current example may lead to unnecessary discussion.

PR #4535, opened 2025-07-31
What's new August

kndehaan:
What's new for deploy August.


mabijkerk:
@kndehaan I made some minor textual changes. Most importantly though, I changed the style of the Upcoming changes to be a bit more subtle. I am satisfied with the overall result. However, I'm not entirely sure about the "What does this mean for existing scenarios?" dropdown. What do you think? How might we improve it?

Screenshot 2025-08-05 at 09 28 13

PR #4546, opened 2025-09-05
Dataset update banner

noracato:
And the one for this scenario interface!

Screenshot 2025-09-05 at 14 13 57

Goes with quintel/my-etm#157


KoenvanB (approved):
Looks good to me!

PR #4549, opened 2025-09-10
Municipal dataset update 2023

KoenvanB:
This PR updates the areas locales for the provinces and RES-regions

- add PV21_fryslan
- add RES regions based on official area codes (ES01 etc)
- add new municipalities since 2019
- rename PV27_2022_noord_holland to "Noord-Holland (2022)"

Please check whether the list of regions in the front end is complete without noticeable errors.


mabijkerk:
This PR also needs to be rebased. I'm for example seeing that an area code for the `nl2023` is added. This has already been done on master. I would keep "Netherlands" as locales, not "Netherlands (2023)".

KoenvanB:
> This PR also needs to be rebased. I'm for example seeing that an area code for the `nl2023` is added. This has already been done on master. I would keep "Netherlands" as locales, not "Netherlands (2023)".

Done!

mabijkerk (approved):

Issue #4548, opened 2025-09-08
Electricity demand per hour chart should be sorted on variability


mabijkerk:
This issue is suggested by @MartLubben. The series can't be sorted dynamically, but some series are typically more variable than others. By placing the most variable series on top, the charts becomes easier to read.

<img width="1260" height="1140" alt="Image" src="https://github.com/user-attachments/assets/342200b9-f71c-4c78-bb65-8a5bdbed8034" />

His suggestion is to use the following distribution:
- Flexibiliteit productiebeperking
- Flexibiliteit Export
- Flexibiliteit Conversie en opslag
- Huishoudens warmtevraag
- Huishoudens heet water
- Huishoudens Overig
- Gebouwen warmtevraag
- Gebouwen overig
- Transport - Elektrische elektrische auto's
- Transport - Overig
- Industrie - Chemie
- Industrie - OVerig
- Land en tuinbouw
- Verliezen
- Overig

I would also suggest to make the series 'clickable'. This is done by changing the chart type to `hourly_stacked_area`. @kaskranenburgQ or @kndehaan perhaps either of you could pick this up coming sprint?

PR #4554, opened 2025-09-18
Visual changes to electricity demand chart

mabijkerk:
This PR addresses #4548, by making two changes to the electricity demand chart:

- It changes the chart type to hourly stacked area in order to make the labels clickable
- It reorders the series from typically more baseload to typically more variable

The only thing I still would like to do is to change the colours of the series. Now that they are reordered, the chart is not as easily to read.

@kaskranenburgQ when reviewing this, can you double-check that I didn't delete any series while reordering the chart?

### DIFF
--- a/config/interface/output_element_series/dynamic_demand_curve.yml
+++ b/config/interface/output_element_series/dynamic_demand_curve.yml
@@ -1,7 +1,7 @@
 ---
 - label: baseload_demand
   color: "#CC0000"
-  order_by: 1
+  order_by: 100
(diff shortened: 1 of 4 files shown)


mabijkerk:
@kaskranenburgQ I intended only to change the label. The query is related to distribution losses, not to storage losses. That is why I think the label was confusing. The underlying query should not have changed.

kaskranenburgQ (approved):
LGTM

PR #4578, opened 2025-10-30
Carbon modelling minor fixes

kndehaan:
This PR implements minor fixes in carbon modelling, specifically:

- Bug fix in final demand international transport chart (solve duplicate query names)
- Order of flows in Liquid fuels sankey
- Locales improvements

Goes with https://github.com/quintel/etsource/pull/3342


mabijkerk:
LGTM. I do have a concern about HVO. With the comments in the description the difference between HVO biodiesel and HVO biokerosene is clear, but based on the sliders themselves it still somewhat confusing to me:

image

Can't we just say "Hydrotreatment to biodiesel" and "Hydrotreatment to biokerosene"?

PR #4584, opened 2025-11-03
Remove ESDL features and set a temporarily unavailable page

noracato:
Due to security issues I turned off the ESDL container

If we want to reactivate the ESDL app, we should also port the old functions that were in ETModel to myETM.

I will create an issue in myETM that will reference this PR for the future, so we will be able to copy the logic.

I kept up the old page, but with a temporarily unavailable message:

Screenshot 2025-11-03 at 17 20 19


mabijkerk (approved):
Nice, I can't speak for the technical changes, but in terms of content it's looking good!

PR #4586, opened 2025-11-06
Filter empty series from hourly_stacked_area chart legend.

### LINKED ISSUE #3839: Series should disappear from the hourly_stacked_area charts when it is empty or only zeros
The standard visualisation of the new flex chart `hourly_flexiblity_electricity` on the `price-senstive-flex` branch is to show all possible series (all that are defined in the output elements yml) in the chart legend, even when a series is not present in the scenario. For example, for a new II3050 National Governance scenario, which does not have installed capacity for nuclear power plants, the series still show in the legend:

Screenshot 2021-09-21 at 10 39 56

@antw is this something that can be changed? I think this is a general issue for the `hourly_stacked_area` chart type because it is also the case for the `merit_order_hourly_flexibility` chart.

aaccensi:
## Description

The `hourly_stacked_area` chart (coffescript route) now also filters any empty series from the legend, same as with the `merit_order_hourly_supply` (non-coffescript route).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement
- [ ] Documentation

## Checklist

## Related Issues

Closes #3839

### DIFF
--- a/app/javascript/charts/HourlyStackedArea.js
+++ b/app/javascript/charts/HourlyStackedArea.js
@@ -34,13 +34,18 @@ export default class HourlyStackedArea extends HourlyBase {
 
   draw() {
     super.draw();
-    this.drawLegend(this.series, 2);
+    this.drawLegend(this.getLegendSeries(), 2);
 
     // Add a "negative region" which shades area of the chart representing values below zero.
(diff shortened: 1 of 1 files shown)


louispt1 (approved):
LGTM

PR #4597, opened 2025-11-19
Rename whats new to releases and ensure redirects to releases page

louispt1:
## Description
This PR goes with [this docs PR to add the releases to the docs](https://github.com/quintel/documentation/pull/266).

This PR renames `whats new` to `releases` wherever possible, and ensures any attempts to visit the whats new will be redirected to the releases page.

One open question is whether we want to keep two links to the docs on the Model homepage:
image

Another consideration is whether we should support the releases/whats new endpoint, or allow it to fail to encourage people to use the new route.

Finally, I noticed we do have a banner that can be retired after a certain date (it has not been used since 2019). We could use that banner to show a sneak peak at the latest changes for a week after releases, for example, if you think that would satisfy the 'brief updates on the ETM' requirement @kndehaan? It looks like this with the text just in a locale:

image

Would you also mind checking the NL locales in this PR for me @kndehaan?

### DIFF
--- a/app/assets/stylesheets/_root.sass
+++ b/app/assets/stylesheets/_root.sass
@@ -56,17 +56,6 @@ body#root
         grid-column-start: 1
         grid-column-end: span 3
         grid-row-start: 1
-        display: flex
-        flex-direction: columns
-        gap: 0.75rem
-        .dates
-          white-space: nowrap
-          background: $landing-background-color
-          padding: 0.5rem 1rem
-          border-radius: 5px
-          height: fit-content
-          p
-            margin-top: 2px
       &.multi-year-charts
         background-image: asset-url("pages/home/multi_year_charts.svg")
         grid-column-start: 4
--- a/app/views/pages/root.html.haml
+++ b/app/views/pages/root.html.haml
@@ -22,17 +22,17 @@
     %h4.title= t('intro.features.title')
   #other-features
     .feature.releases
-      %div
-        %h4= t('intro.releases.title')
+      %h4= t('intro.releases.title')
 
-        %p= t('intro.releases.description')
-        %p
-          = link_to(releases_url, target: '_blank', rel: 'noopener noreferrer') do
-            = t('intro.releases.link')
-            →
-      .dates
-        %p= t('intro.releases.date')
-        %span= l(ETModel::Version::DATE, format: :long)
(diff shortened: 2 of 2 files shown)


kndehaan:
Here already my remarks on your points above, I will look into the changes more deeper next. 

> One open question is whether we want to keep two links to the docs on the Model homepage

I think you refer here to the link to releases and the link to documentation, right? I think we should keep them both since they represent different things that are both relevant on the landing page: the documentation on how to use the ETM and releases for seeing changes to the model, which just now happens to live in the same interface as the documentation. 

> Another consideration is whether we should support the releases/whats new endpoint, or allow it to fail to encourage people to use the new route.

I don't have a strong opinion on this. I think people generally don't use the old link to the what's new themselves but go to the What's new page via the link on the landing page or the technical newsletter. 

> We could use that banner to show a sneak peak at the latest changes for a week after releases, for example, if you think that would satisfy the 'brief updates on the ETM' requirement @kndehaan? 

I'm not in favor of using a banner to indicate that changes have been made to the model, in my opinion there's no need to give it that much attention. I would be in favor to highlight the most important changes in 1-2 sentences in the releases block though:
image

That is what we now do with the current What's new setup. So that it would look something like this:

image

kndehaan (approved):

kndehaan:
I changed the definitions on the landing page to "Updates" instead of "Releases". 

I'm don't know how we can neatly implement the data of the latest release in the box. Just putting it in the text doesn't veel nice to me. @noracato do you have any ideas on how we could easily do this? Putting something in the right corner perhaps?

Screenshot 2025-11-25 at 11 30 45

kndehaan:
> For sure, I will make the changes so it opens in a new tab

Coool, I see that the Releases page opens in a new tab now. Is this possible for the other pages as well? (Collections, Dataset Manager, Documenation, API, Quintel website) Alternatively, I could open a new issue for this...

image

kndehaan:
Amazing! Ok now we're really good to go!

mabijkerk:
I like the idea of having the last update noted but in terms of UI it looks a bit 'added on' for me. Is there a way to visually integrate it better in the existing box?

Something more like:

image

louispt1:
Bold or just like this? @mabijkerk 

image

I agree maybe having it more 'inline' looks cleaner

PR #4687, opened 2026-03-26
Replace scenario report retirement banner by a friendly notice

aaccensi:
#### Context

We recently replace the scenario report by a notice explaining both Scenario Report and Energy Mix are retired but the message we wanted to convey of collaboration was lost.

#### Implemented changes

We replace the yellow banner by a friendlier notice paragraph and changed the spacing and wording to better convey the real situation in hopes people will actually read the content and not skip straight to the example charts.

## Checklist


mabijkerk (approved):
Very nice!

PR #4696, opened 2026-04-08
Verify database set up or throw warning

louispt1:
#### Context
Adds a warning when the databases do not set up properly using the databases instructions based on MyETM's readme.

#### Related
Goes with [this myetm PR](https://github.com/quintel/my-etm/pull/238)
Goes with [this etengine PR](https://github.com/quintel/etengine/pull/1733)

## Checklist

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -30,6 +30,18 @@ FileUtils.chdir APP_ROOT do
   puts "\n== Preparing database =="
   system! "bin/rails db:prepare"
 
+  puts "\n== Verifying database setup =="
+  verify_cmd = <<~BASH
+    mysql -u root -e "SELECT COUNT(*) FROM sessions LIMIT 1;" etmodel_dev 2>&1 > /dev/null
+  BASH
+
+  if system(verify_cmd)
+    puts "✓ Database setup verified successfully"
+  else
+    puts "⚠ Warning: Database may not be fully set up. Tables might be missing."
+    puts "Try running: bin/rails db:drop db:create db:schema:load db:seed"
+  end
+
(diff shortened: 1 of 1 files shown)


bin/setup
```diff
@@ -30,6 +30,18 @@ FileUtils.chdir APP_ROOT do
...
+    mysql -u root -e "SELECT COUNT(*) FROM sessions LIMIT 1;" etmodel_dev 2>&1 > /dev/null
+  BASH
+
+  if system(verify_cmd)
+    puts "✓ Database setup verified successfully"
+  else
+    puts "⚠ Warning: Database may not be fully set up. Tables might be missing."
+    puts "Try running: bin/rails db:drop db:create db:schema:load db:seed"
+  end
+
```
aaccensi, line 44: If I understand the intent of these changes correctly, I believe there might be a better check.

Current check only fails on completely missing table, so perhaps is better to use a more rails native way of validating the db and its migrations, that in addition does not assume that mysql is installed even though that is sort of a hard requisite.

The following is not necessary the final solution but an idea:
```
puts "\n== Verifying database setup =="
if system("bin/rails runner 'ActiveRecord::Migration.check_pending!' 2>/dev/null")
  puts "✓ Database migrations are up to date"
else
  puts "⚠ Warning: Database may not be fully set up. Migrations may be pending or the schema is missing."
  puts "Try running: bin/rails db:drop db:create db:schema:load db:seed"
end

```

aaccensi (commented):
I think the intention is good but perhaps we need to bring it a step further, let me know what you think?

aaccensi (approved):
Alright, as a quick fix sounds good!

PR #4698, opened 2026-04-08
Add must-run gas plants

claudiavalkenier:
#### Context
For a certain project, the goal is to introduce two must-run gas-fired electricity plants, based on existing dispatchable gas-fired electricity plants. These are the CCGT gas plant and the gas turbine plant.

This means that the current CCGT gas plant and gas turbine plants will be renamed to dispatchable, including all related dependencies. Subsequently, duplicates of both plant types will be created and redefined as must-run.

Initially, this will only be implemented for the future year, meaning that no additional data needs to be read from the dataset.

#### Implemented changes
- Rename sliders
- Add must-run sliders
- Specify slider groups
- Update graphs and tables

#### Related

Goes with pull requests:
- ETModel: https://github.com/quintel/etmodel/pull/4698
- ETSource: https://github.com/quintel/etsource/pull/3433
- ETEngine: https://github.com/quintel/etengine/pull/1734
- Documentation: https://github.com/quintel/documentation/pull/306
- Mechanical Turk: https://github.com/quintel/mechanical_turk/pull/217
- Multi-year-charts: https://github.com/quintel/multi-year-charts/pull/119
- ETDataset: https://github.com/quintel/etdataset/pull/1083

## Checklist

Tagging @kndehaan as functional reviewer!


config/interface/output_element_series/merit_order_hourly_supply.yml
kndehaan, line 1: Could you apply different colours for the must-run / dispatchable variants?

config/interface/output_element_series/merit_order.yml
kndehaan, line 1: Can you also apply different colours here for must-run / dispatchable variants? Just a slightly darker or lighter (not a totally different colour)

kndehaan (commented):
Locales of must-run and dispatchable is missing in this chart:

Image

The CCS variant should also have "(dispatchable)" in its label in the merit order chart table:

Image

The must-run plants need to be added in the economic viability table, the labels of the dispatchable ones still need to be updated. Furthermore, some values are missing for the gas turbine and CCS technology, I think something goes wrong there:

Image

kndehaan:
There's an additional step needed that I didn't think of before: we need to rename the node source analyses on ETDataset and copy them for the must_run plants as well. Could you therefore make a branch on ETDataset as well to do this? 

Additionally, the node source analyses should be added / renamed in the ETModel file `nodes_download_source.yml` (this way, users can download the excels under "Technical and financial properties" of the technologies.

kndehaan (approved):
Looking good!

PR #4715, opened 2026-05-12
Remove Sentry browser profiling

aaccensi:
#### Context

Sentry browser profiling was enabled but proved mostly unusable. The call tree was dominated by minified framework code even with source maps in place. We're paying for profiling hours with little actionable insight.

#### Implemented changes

- Removed `browserProfilingIntegration` from Sentry init
- Removed `profileSessionSampleRate` and `profileLifecycle` options
- Removed `sentry_profiles` from the globals passed to the frontend and from `settings.yml`

## Checklist


noracato (approved):
Thanks! LGTM

PR #4719, opened 2026-05-29
Add direct emissions method with data export

kndehaan:
#### Context

This PR adds a new direct emissions method to the model, together with a detailed data export containing all greenhouse gas emissions of a scenario. 

#### Implemented changes
- New engine methods and node attributes that calculate relevant direct emissions, based on presence of (molecule) node group `emissions`
- Data processing pipeline that generates start year emissions data for country datasets, based on reported emission data from UNFCCC. Start year data is added as `emissions.csv` file to ETSource using a rake task. 
- New molecule nodes that read static emission data from `emissions.csv`, specified for co2 and other_ghg
- Detailed data export containing relevant direct reporting emission results for all relevant nodes, and categorises these into ETM sector and subsector. 

#### Related

Goes with pull requests:
- https://github.com/quintel/etengine/pull/1754
- https://github.com/quintel/atlas/pull/189
- https://github.com/quintel/etsource/pull/3444
- https://github.com/quintel/etmodel/pull/4719
- https://github.com/quintel/etdataset/pull/1087
- https://github.com/quintel/etlocal/pull/695
- https://github.com/quintel/documentation/pull/321

## Checklist

### DIFF
--- a/config/interface/slides/data_export.yml
+++ b/config/interface/slides/data_export.yml
@@ -11,6 +11,10 @@
   position: 105
   sidebar_item_key: data_export
   output_element_key: sankey_energy_overview
+- key: data_export_emissions
+  position: 107
+  sidebar_item_key: data_export
+  output_element_key: co2_emissions
 - key: data_export_molecule_flows
   position: 110
   sidebar_item_key: data_export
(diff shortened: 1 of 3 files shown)


mabijkerk:
Technically no comments. Some thoughts on content:

- I think that we should state that this method is aligned with standard reporting framework, either here or in the docs.
- Should we more clearly state that this is a `beta` release functionality? Then we can state which things are missing: the other GHG for Dutch regions, the chart shown is still based on primary emissions etc.

kndehaan:
I processed your first point in the docs (PR will created today). I agree with your second point and I think that the docs would be a good location for this. As well as in the release notes. I will propose something for this in the docs.

mabijkerk (approved):

PR #4723, opened 2026-06-10
Clear Solid Cache on deploy and expire by_slide cache

### LINKED ISSUE #123: Gas power plants missing
#### Context
On ETModel there is a list of gas power plants that looks like this:

Image

In the input settings expanded view from the Collections some appear to be missing:

Image

#### Steps to reproduce

On pro, latest, using nl2023 blank transition path.

#### Proposed solution
Do not know the proposed solution yet.

louispt1:
#### Context
Four gas power plant sliders (must-run + dispatchable CCGT and turbine) were missing from the Collections input settings view. The deploy hook was not updated when we moved to solid cache, meaning that the cache entries were surviving deploys and becoming outdated with respect to the actual data.

##### Changes:
- The deploy `clear_cache` hook now runs `Rails.cache.clear` on the app server 
- The `by_slide` response now expires after 1 day as a fallback
- Removed the dead memcached Capistrano rake tasks

Before:
image

After:
image

#### Related
Closes [this MYC issue](https://github.com/quintel/multi-year-charts/issues/123)

## Checklist


noracato (approved):
Awesome, thanks for this fix and removing some more old capistrano stuff

PR #4735, opened 2026-06-23
Completely retire the investment table

### LINKED ISSUE #3781: Investment table should be critically reviewed for outdated and missing nodes
In issue #3780 it turned out that some locales were missing from the investment table. These have since been added, but during that process the following points came up:

Screenshot 2021-07-29 at 14 49 18

- The table includes blast furnace for steel, but not the blast furnace BAT, or the cyclone oven
- The table includes steel investment costs as separate technologies, but for aluminium, which is also based on different technologies, it only includes the aluminium network gas burner
- Wind turbines for hydrogen production do not seem to be included in the table
- Solar panels for hydrogen production do not seem to be included in the table

Based on the points addressed above it is clear that the table should be critically reviewed. All in all, it shows that this table is not as structurally maintained as could be, perhaps we could write a spec for this @michieldenhaan?

### LINKED ISSUE #3930: Should we bring back the investment table?
The investment table was hidden from the front-end because it was broken and removed from the scenario report, see #3911. The table would be replaced by a new data export during the cost project. The new data export contains information about the CAPEX and OPEX (including and excluding CCS), the number of units and the total input and output of the node. 

It is not possible anymore to compare the total investment in the base year with the total investment in the end year. Do we want to repair the investment table to give users access to this information in the front-end? @quintel/core, @DorinevanderVlies and I would like to hear your opinion about this.

mabijkerk:
#### Context
The investment table has been hidden for a long time, since it has been replaced by the costs data export. The queries still lingered in ETSource and were kept (badly) up to date.

#### Implemented changes
To avoid having to keep redundant queries up to date, the investment table is now deleted. It can also be returned in the future based on this pull request.

#### Related

Closes issue https://github.com/quintel/etmodel/issues/3930
Closes issue https://github.com/quintel/etmodel/issues/3781

Goes with https://github.com/quintel/etsource/pull/3451 since that PR showed that there were broken queries in the investment table queries.

## Checklist


kndehaan (approved):
Looking good!

PR #4747, opened 2026-07-06
SSO: Login/Logout

louispt1:
#### Context
Adopt shared JWT session cookie for authentication, retire per-app OAuth session

Mirrors etengine changes:
- Replaced the local acess-token/session-based sign-in with the domain-wide `etm_session` cookie minted by MyETM.
- current_user resolves via User.from_jwt! against the cookie claims
- `ETModel::Tokens` is replaced with `Identity::TokenDecoder` (unified logic in the identity gem). Other superceded methods also retired or unified into the gem for etengine and etmodel
- Adds `session_access_token/session_access_token_expires_at` helpers so `my_etm_client/engine_client` and the front-end can use the cookie's bearer token and know when to refresh, and `app.coffee` gains client-side refresh  scheduling against that expiry.

#### Related - Goes with the other SSO: Login/Logout PRs:
[identity rails](https://github.com/quintel/identity_rails/pull/8)
[etengine](https://github.com/quintel/etengine/pull/1776)
[etmodel](https://github.com/quintel/etmodel/pull/4747)
[myetm](https://github.com/quintel/my-etm/pull/276)
[collections](https://github.com/quintel/multi-year-charts/pull/124)
[ETLauncher](https://github.com/quintel/ETLauncher/pull/2)

## Checklist
- [x] I have tested these changes --> **NOTE: These changes were tested on the ETLauncher bridge network. Testing the federated cookie setup on the standard dev setup will not work because cookies cannot be federated across localhost as a domain**
- [x] I have updated documentation as needed (comments only)


noracato (approved):
Also very happy here to see so much code go. It feels a lot safer without using any globals.

aaccensi (approved):
Looks good!

PR #4750, opened 2026-07-14
Remove broken Saved Scenario Reports feature

### LINKED ISSUE #4669: Cleanup of old/broken Saved Scenario Reports feature
#### Context

During the removal of Energy Mix infographic and adding the deprecation notice to the Scenario Report feature, some queries were found that looked only related tot he Energy Mix yet under a completely different feature. 

That feature that generated a CSV export for a saved scenario when those were still in etmodel. It is broken, likely since saved scenarios were moved to my-etm.  

#### Proposed enhancement

Since there are no complains about this feature and it has been broken a very long time it is safe to remove, for which the following actions have been estimated:

To delete the following files:
File | Notes
-- | --
app/controllers/saved_scenario_reports_controller.rb | The broken controller
app/views/saved_scenario_reports/show.csv.erb | CSV view template
app/services/fetch_api_scenario_queries.rb | Only used by this controller
spec/services/fetch_api_scenario_queries_spec.rb | Spec for the above service
config/saved_scenario_reports/ovm.yml | The only report config (delete dir too)

To remove the following line from the routes: `config/routes.rb:55 `
```
Remove get '/report/:report_name' => 'saved_scenario_reports#show'
```

#### Success criteria

Remove any trace of this old feature without disrupting anything else.

aaccensi:
#### Context

While removing the Energy Mix infographic and adding the deprecation notice to the Scenario Report feature, some queries turned up that belonged to a different, forgotten feature: a CSV export for saved scenarios (/saved_scenarios/:id/report/:report_name) driven by a YAML query template. It dates from when saved scenarios still lived in etmodel and has been broken since they moved to MyETM. It has been broken for a long time with no complaints, so this PR removes it entirely rather than fixing it.

#### Implemented changes

* Removed `SavedScenarioReportsController` (broken CSV report controller)
* Removed the CSV view template `app/views/saved_scenario_reports/show.csv.erb`
* Removed the `FetchAPIScenarioQueries` service and its spec (only used by this controller)
* Removed `config/saved_scenario_reports/ovm.yml`, the only report config (directory deleted too)
* Removed the route `get '/report/:report_name' => 'saved_scenario_reports#show'` from `config/routes.rb`
* Removed the `Engine::Query` model — orphaned once the service was deleted (its sole consumer)

#### Related

Closes quintel/etmodel#4669

## Checklist

- [X] I have tested these changes
- [X] I have tagged the relevant people for review


app/models/engine/query.rb
noracato, line 1: This was just orphaned anyways right?

noracato (approved):

app/models/engine/query.rb
aaccensi: yes, this was just old orphaned files and a route

PR #4774, opened 2026-08-11
Cherry-picks the shared-session SSO work from master onto 2025-01

louispt1:
#### Context
It is necessary to apply some of the SSO changes on master to the stable version so that single login and logout works for the stable version as well as on latest.

#### Implemented changes
- I verified the cherry-picked changes end to end with ETLauncher: I checked that the sign in sets the cookies, signs in everywhere (once) and signs out everywhere (once). The token refreshes correctly.

The auth context here now closely resembles master.

One additional change was required here: the Dockerfile now specifies `libv8-node` and `mini_racer` before `bundle install` as otherwise etlauncher could not build the image.

#### Related
[etengine 2025-01 cherry pick](https://github.com/quintel/etengine/pull/1798)
[etmodel 2025-01 cherry pick](https://github.com/quintel/etmodel/pull/4774)
[collections 2025-01 cherry pick](https://github.com/quintel/multi-year-charts/pull/126)

## Checklist


aaccensi (dismissed):
I think there is a small gap for a file that has no counterpart on master and therefore cannot be fixed by the cherry-pick alone.

`app/helpers/energy_mix_helper.rb:10` still calls `identity_access_token`, which no longer exists in identity_rails. `layouts/energy_mix.html.haml:16` calls the helper unconditionally, so `/scenarios/:id/energy_mix `raises NoMethodError → 500 for signed-in users while guests take the nil branch and the page renders fine.

I beleive the fix is exactly the same one as in `app/helpers/pages_helper.rb:59`:
```
{ endpoint: url, token: session_access_token }
```

louispt1:
> I think there is a small gap for a file that has no counterpart on master and therefore cannot be fixed by the cherry-pick alone.
> 
> `app/helpers/energy_mix_helper.rb:10` still calls `identity_access_token`, which no longer exists in identity_rails. `layouts/energy_mix.html.haml:16` calls the helper unconditionally, so `/scenarios/:id/energy_mix `raises NoMethodError → 500 for signed-in users while guests take the nil branch and the page renders fine.
> 
> I beleive the fix is exactly the same one as in `app/helpers/pages_helper.rb:59`:
> 
> ```
> { endpoint: url, token: session_access_token }
> ```

Great catch! I just pushed the suggested fix 👍

PR #4796, opened 2026-08-27
Households insulation chart

jort-wolda:
#### Context
The etsource changes for the charts for the households - insulation section.

#### Implemented changes
- Added a filter functionality for category bars
- Added a new chart: residential_heating_demand_per_house.yml
- Made some changes to the existing chart: residential_heating_demand.yml
- Set the new chart as the default at the households/insulation slide

#### Related
Goes with pull requests:
- https://github.com/quintel/etsource/pull/3479

## Checklist

- [ ] I have updated documentation as needed (I don't think documentation is needed for this)


robindesmit-quintel:
Review:
- Behaviour> per person > hot water: _"these changes in energy demand do not depend on the technologies chosen below (Hot water slide)"_ The technologies are not chosen below but in the "space heating & hot water" slide! This is the case for all sliders in `Behaviour.`
_"Here you indicate how you expect demand for hot water per person to change. This is mainly determined by growth in prosperity."_ I would say it can depend on many things: the weather, the price of hot water, etc. Other sliders include more of these situations, should we put that in the description of `Behaviour` itself?
- Behaviour: why indicate some sliders the change per year and other the total change from 2023 to 2050? Maybe we should add a note about this. 
- How extensive should the explanation be of the sliders? The explanation of Cooling> ground heat pump is very long, might be better included in the documentation? The explanation of sliders in Insulation is the same for all sliders, might be better included in the documentation?

kndehaan:
@robindesmit-quintel thanks for the review notes. As Jort's work in this PR is only related to adding new user output (charts), I think it's not needed to put too much effort in improving (existing) slider texts within this PR.

I'd suggest to only pick up the following from Robin's review comments:
> - Behaviour> per person > hot water: "these changes in energy demand do not depend on the technologies chosen below (Hot water slide)" The technologies are not chosen below but in the "space heating & hot water" slide! This is the case for all sliders in Behaviour.

config/locales/interface/output_element_series/nl_labels.yml
```diff
@@ -1425,6 +1435,7 @@ nl:
         sustainable_ammonia: "Hernieuwbare ammoniak"
         television: "Televisie"
         target: "Doel"
+        terraced: "Rijtjes woningen"
```
kndehaan: This label deviates from the slider name. Could you changes this to "Rijtjeshuizen"?

Image

config/interface/output_element_series/residential_heating_demand.yml
kndehaan, line 1: @aaccensi do you understand why there's a different order of the legend labels compared to the series in the chart? Do order of the chart is logical (newly built at the top, oldest at bottom). The label order seems quite random though. 

Image

config/locales/interface/output_elements/en_demand.yml
```diff
@@ -493,6 +493,10 @@ en:
       title: Residential space heating demand
       description: |
         This bar chart shows the useful demand for space heating per residence type and building period.
+    residential_heating_demand_per_house:
+      title: Residential space heating demand per house
+      description: |
+        This bar chart shows the useful demand for space heating per house, per building period and residence type.
```
kndehaan: - Could you change the title to ".. per residence". 
- The "..per house, ... residence type" seems somewhat double. 
- Could you rename "building period" to "construction period"

config/interface/slides/households.yml
```diff
@@ -12,7 +12,7 @@
   key: demand_households_insulation
   position: 105
   sidebar_item_key: households
-  output_element_key: residential_heating_demand
+  output_element_key: residential_heating_demand_per_house
```
kndehaan: Could you rename the output_element_key to ...per_residence ?

config/locales/interface/output_element_series/nl_labels.yml
```diff
@@ -1,6 +1,10 @@
 nl:
   output_element_series:
     labels:
+        '1945_1964': "1945 - 1964"
+        '1965_1984': "1965 - 1984"
+        '1985_2004': "1985 - 2004"
+        '2005_present': "2005 - present"
```
kndehaan: Should be "2005 - heden"

config/locales/interface/output_elements/en_demand.yml
jort-wolda: I understand that it sounds double but I don't mean the exact same thing. The first 'per house' mens that the unit is GJ/residence, not GJ of the total building stock.

The second means that there are different values for different construction periods and different residence types.

I'll try to reword so that it is more clear

config/interface/output_element_series/residential_heating_demand_per_residence.yml
kndehaan, line 1: This file name itself should still be renamed to `residential_heating_demand_per_residence.yml`

kndehaan (approved):
Looking good!

PR #4798, opened 2026-08-28
Heat demand per building and 'flipped' chart

jort-wolda:
#### Context
Due to the new slider set-up for the built environment the charts are also updated. 

In the PRs below a new chart is added that shows heat demand per residence:
- https://github.com/quintel/etmodel/pull/4796
- https://github.com/quintel/etsource/pull/3479

In this PR some additional changes are made to heat/insulation charts.

#### Implemented changes
- Added a heat demand per building chart for residential buildings
- Added a new residential heat demand chart that shows the totals with the construction periods on the y-axis instead of the residence types
- Renamed the existing residential heat demand chart to indicate the difference from the new chart
- Changed the description on the insulation slide to mention the totals charts.

#### Related
Goes with pull requests:
- https://github.com/quintel/etsource/pull/3482

## Checklist

### DIFF
--- a/config/interface/output_element_series/buildings_heating_demand_per_building.yml
+++ b/config/interface/output_element_series/buildings_heating_demand_per_building.yml
@@ -1,0 +1,25 @@
+---
+- label: buildings_present
+  color: "#A6A6A6"
+  order_by: 10
+  group: buildings_present
+  show_at_first:
+  is_target_line: false
+  target_line_position:
+  gquery: useful_demand_buildings_present_per_building
+  is_1990:
+  dependent_on:
+  output_element_key: buildings_heating_demand_per_building
+  key: useful_demand_buildings_present_per_building_buildings_heating_demand_per_building
+- label: buildings_future
+  color: "#FFB000"
+  order_by: 20
+  group: buildings_future
+  show_at_first:
+  is_target_line: false
+  target_line_position:
+  gquery: useful_demand_buildings_future_per_building
+  is_1990:
+  dependent_on:
+  output_element_key: buildings_heating_demand_per_building
+  key: useful_demand_buildings_future_per_building_buildings_heating_demand_per_building
--- a/config/interface/output_element_series/buildings_heating_demand_per_building.yml
+++ b/config/interface/output_element_series/buildings_heating_demand_per_building.yml
@@ -6,7 +6,7 @@
   show_at_first:
   is_target_line: false
   target_line_position:
-  gquery: useful_demand_buildings_present_per_building
+  gquery: useful_demand_buildings_present_per_building_bar
   is_1990:
   dependent_on:
   output_element_key: buildings_heating_demand_per_building
@@ -18,7 +18,7 @@
   show_at_first:
   is_target_line: false
   target_line_position:
-  gquery: useful_demand_buildings_future_per_building
+  gquery: useful_demand_buildings_future_per_building_bar
   is_1990:
   dependent_on:
   output_element_key: buildings_heating_demand_per_building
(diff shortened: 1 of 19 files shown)


config/interface/output_element_series/buildings_heating_demand_per_building.yml
kndehaan, line 1: The chart now shows "GJ/WEQ" when viewing the model in English. WEQ however does not have a meaning in English and is very Dutch. We might need to discuss this, since residence equivalents might be difficult to abbreviate. Perhaps best option is to stick with GJ/# and add a sentence in the chart description that "#" is residence equivalents / WEQ. Similar to what is described in the slider texts for number of buildings. @jort-wolda what do you think?

Image

kndehaan: Additionally, can we set this chart as the default to be shown at the Insulation slide? And if we do so, we could also provide in the slide text the link to the chart with total heat demand for buildings (`buildings_heating_demand.yml`) (similar to the Insulation slide text for households).

config/locales/interface/slides/en_demand.yml
```diff
@@ -180,6 +180,10 @@ en:
         The sliders below set the reduction in space heating demand due to improved insulation. Go
         to the section <a href="/scenario/demand/households/behaviour">behaviour</a>
         to set changes in space heating demand due to behavioural changes.
+        <br/><br/>
+        The default chart shows the space heating demand per residence. You can also view the space heating demand of the total housing stock,
+        either by <a href="#" class="open_chart" data-chart-key="residential_heating_demand_by_type" data-chart-location="side">residence type</a>
+        or by <a href="#" class="open_chart" data-chart-key="residential_heating_demand_by_period" data-chart-location="side">construction period</a>.
```
kndehaan: Try to avoid using "you / je" in writing slide, slider, etc. texts (also see writing [guidelines](https://docs.energytransitionmodel.com/contrib/authoring-texts#to-be-avoided)).

config/interface/output_elements/demand_households.yml
```diff
@@ -196,7 +196,29 @@
...
+  max_axis_value:
+  min_axis_value:
+  hidden: false
+  requires_merit_order: false
+  dependent_on: ''
+  output_element_type_name: category_bar
+  config:
+    show_total_row: true
+    serie_selections:
+    - bar
```
kndehaan, line 209: Why is this called "bar"? Shouldn't something like "all" or "all_residences" be more intuitive?

config/interface/output_element_series/buildings_heating_demand_per_building.yml
jort-wolda: All done:
- set this as the default chart
- changed slider descriptions for both buildings and households to refer to the other charts. Deleted this from the slide text
- due to the addition the slider texts became very long. I changed the wording to shorten it a bit again.
- Changed the GJ/WEQ to GJ/# and added explanation to the chart description about WEQs. Also changed the unit in the queries in etsource (was also GJ/WEQ)

config/interface/output_elements/demand_households.yml
jort-wolda: The _serie_selections_ function filters based on the name of the queries. All queries have different names so the only way to include them is to pick a piece of text that they all have. In this case I picked bar, another option might have been useful_demand. 

Something like "all" wouldn't work since it isn't part of the query names. If we'll make such a chart in the future we can keep this in mind when naming the queries.

kndehaan (approved):
Good to go!

